PR #17, opened 2019-06-04
ENH improve robustness to scale for optimization

### LINKED ISSUE #6: Parameter tol in LGCD depends on the scale of z
The `tol` parameter in the LGCD algorithm is dependent on the activation magnitude. It should be normalized to avoid getting many `Regularization parameter reg is too large`.

tomMoral:
The scaling `X / X.std()` is needed to avoid numerical issues in the optimization but it is not really necessary. It would be nice to make the optimization more robust to sucha scaling issue.
For the Z-step, the optimization is not really impacted, only the stopping criterions are. I propose to correctly scale the tolerance parameter in `lgcd` to make it independent of the scale.
For the d/uv-steps, the norm of the gradient might be too low when using unscaled MEG signals. To make is more friendly for practitioners, I implemented a scaling for the optimization inside `learn_d_z_multi`. This does not change the results but makes the optimization more robust. The main issue with this is that it breaks the equivalence between successive `partial_fit` and an online fit if the `std` of the signals are not unitary. But I would say it is not a big issue.

Closes #6

### DIFF
--- a/alphacsc/online_dictionary_learning.py
+++ b/alphacsc/online_dictionary_learning.py
@@ -57,7 +57,7 @@ class OnlineCDL(ConvolutionalDictionaryLearning):
             name="OnlineCDL")
 
     def partial_fit(self, X, y=None):
-        # Successice partial_fit are equivalent to OnlineCDL only if
+        # Successive partial_fit are equivalent to OnlineCDL only if
         # the X passed to this method are taken from a normalized
         # X_full ( X_full / X_full.std())
         self._check_param_partial_fit()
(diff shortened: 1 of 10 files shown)


alphacsc/online_dictionary_learning.py
```diff
@@ -57,6 +57,9 @@ def __init__(self, n_atoms, n_times_atom, reg=0.1, n_iter=60, n_jobs=1,
             name="OnlineCDL")
 
     def partial_fit(self, X, y=None):
+        # Successice partial_fit are equivalent to OnlineCDL only if
```
agramfort:
```suggestion
        # Successive partial_fit are equivalent to OnlineCDL only if
```

tomMoral:
Yes I am currently trying to update the example but I wanted to fix this bad behavior as it was the main impediment in the mail from Lari. The resulting optimization was never started beause the tolerance was too high for MEG signal. Once I merge this PR, I will get back to updating the example using MNE.

PR #32, opened 2021-04-01
move ci from travis to github actions

hndgzkn:
Uses Github actions instead of travis with Python 3.6. 

1. Some tests do not pass and linting problems for versions >3.6. 
2. No action  added for Github pages. 

We can fix 1 and 2 here or in pull requests.

@tomMoral I am not sure if it is necessary to enable github actions in alphacsc to make ci pipeline run.


agramfort:
@hndgzkn I just gave you admin access on repo so you can setup the GH
Actions

>
thanks !

agramfort:
all green !

thx @hndgzkn

PR #35, opened 2021-04-06
Test for python3.8

hndgzkn:
Update workflows to use Python 3.8.
Fix problems while testing with Python3.8
Fix linting for Python3.8

Tests run without problems for 3.7 as well.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python: [3.6]
+        python: [3.8]
     steps:
       - uses: actions/checkout@v2
       - uses: conda-incubator/setup-miniconda@v2
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -25,10 +25,10 @@ jobs:
           conda --version
           which python
           pip install numpy cython
-          pip install -e .
           pip install pytest">=3.6" coverage pytest-cov coveralls
       - name: Run unit tests
         run: |
+          pip install -e .
           pytest --cov=alphacsc --cov-report=xml
       - name: Upload coverage
         uses: codecov/codecov-action@v1
--- a/alphacsc/datasets/hcp.py
+++ b/alphacsc/datasets/hcp.py
@@ -221,7 +221,7 @@ def make_array(X, equalize='zeropad'):
                            axis=-1) for x in X
         ])
     else:
-        raise ValueError(f"The equalize parameter should be in "
+        raise ValueError("The equalize parameter should be in "
                          f"{'crop', 'zeropad'}, got '{equalize}'.")
 
     return X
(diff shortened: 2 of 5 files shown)


alphacsc/learn_d_z_multi.py
```diff
@@ -145,8 +145,8 @@ def learn_d_z_multi(X, n_atoms, n_times_atom, n_iter=60, n_jobs=1,
     """
 
     assert lmbd_max in ['fixed', 'scaled', 'per_atom', 'shared'], (
-        "lmbd_max should be in {'fixed', 'scaled', 'per_atom', 'shared'}, "
-        "not '{}'".format(lmbd_max)
+        "lmbd_max should be in {'fixed', 'scaled', 'per_atom', 'shared'}"
+        f"not '{lmbd_max}'"
```
agramfort, line 149:
```suggestion
        "lmbd_max should be in {'fixed', 'scaled', 'per_atom', 'shared'}, "
        f"got '{lmbd_max}'"
```

alphacsc/datasets/hcp.py
```diff
@@ -221,7 +221,7 @@ def make_array(X, equalize='zeropad'):
                            axis=-1) for x in X
         ])
     else:
-        raise ValueError("The equalize '{}' is not valid. It should be in "
-                         "{'crop', 'zeropad'}".format(equalize))
+        raise ValueError(f"The equalize '{equalize}' is not valid. "
+                         "It should be in {'crop', 'zeropad'}")
```
agramfort, line 225:
```suggestion
        raise ValueError(f"The equalize parameter should be in "
                         f"{'crop', 'zeropad'}, got '{equalize}'.")
```

agramfort (approved):

alphacsc/datasets/hcp.py
```diff
@@ -221,7 +221,7 @@ def make_array(X, equalize='zeropad'):
                            axis=-1) for x in X
         ])
     else:
-        raise ValueError(f"The equalize '{equalize}' is not valid. "
-                         "It should be in {'crop', 'zeropad'}")
+        raise ValueError(f"The equalize parameter should be in "
```
tomMoral, line 224:
```suggestion
        raise ValueError("The equalize parameter should be in "
```

.github/workflows/unittests.yml
```diff
@@ -25,10 +25,10 @@ jobs:
           conda --version
           which python
           pip install numpy cython
-          pip install -e .
           pip install pytest">=3.6" coverage pytest-cov coveralls
       - name: Run unit tests
         run: |
+          pip install -e .
```
tomMoral, line 31: Not sure why you change this?
If no particular reason, I would keep the `install` in the `Installation` step.

tomMoral (approved):
LGTM, just 2 coments

.github/workflows/unittests.yml
hndgzkn: I thought of installation step as installation of dependencies. 
If you prefer the previous version, I'll revert that change.

tomMoral: Let's keep like this then :)
Not really a big deal.

tomMoral:
Merging, thx @hndgzkn .

Issue #40, opened 2021-09-21
Add numpydoc checks to CI jobs


rprimet:

rprimet:
Related:

- https://github.com/numpy/numpydoc/issues/213
- https://github.com/scikit-learn/scikit-learn/pull/15404/files

PR #41, opened 2021-09-22
Add numpydoc checks

rprimet:
Draft numpydoc checks, not ready yet (we should probably move to a whitelist).

See #40


agramfort:
Maybe we could use pydocstyle with numpy parameter ? No strong opinion

PR #37, opened 2021-04-20
Deploy GitHub pages

hndgzkn:
Adds Github action to build the docs and deploy to github pages.

A token should be generated and added to alphacsc project secrets with name DEPLOY_PAGES.

Addresses issue #34. 

Fixes mne-somato problems mentioned in pull request #28 to be able to run the examples.

PR #44, opened 2021-09-23
Update github pages action to use a more recent sphinx image

### LINKED ISSUE #43: Deploy GitHub pages  job fails
`Deploy GitHub pages` CI job fails due to https://github.com/ammaraskar/sphinx-action/issues/33

hndgzkn:
Fixes #43 

https://github.com/ax3l/sphinx-action is a fork of https://github.com/ammaraskar/sphinx-action that fixes sphinx image version.

### DIFF
--- a/examples/csc/plot_lfp_data.py
+++ b/examples/csc/plot_lfp_data.py
@@ -14,13 +14,28 @@ atoms from Local Field Potential (LFP) data [1].
 ###############################################################################
 # First, let us fetch the data (~14 MB)
 import os
-from mne.utils import _fetch_file
+from mne.datasets import fetch_dataset
+from mne.utils import get_config
 
 url = ('https://github.com/hitziger/AWL/raw/master/Experiments/data/'
        'LFP_data_contiguous_1250_Hz.mat')
-fname = './LFP_data_contiguous_1250_Hz.mat'
-if not os.path.exists(fname):
-    _fetch_file(url, fname)
+
+folder_name = "LFP"
+archive_name = "LFP_data_contiguous_1250_Hz.mat"
+
+fetch_dataset(
+    {"dataset_name": "LFP_data",
+     "url": url,
+     "archive_name": archive_name,
+     "folder_name": folder_name,
+     "hash": None
+     },
+    path=None,
+    force_update=False
+)
+
+fname = os.path.join(get_config("MNE_DATA"),
+                     "LFP", "LFP_data_contiguous_1250_Hz.mat")
 
 ###############################################################################
 # It is a mat file, so we use scipy to load it
--- a/examples/csc/plot_lfp_data.py
+++ b/examples/csc/plot_lfp_data.py
@@ -23,7 +23,7 @@ url = ('https://github.com/hitziger/AWL/raw/master/Experiments/data/'
 folder_name = "LFP"
 archive_name = "LFP_data_contiguous_1250_Hz.mat"
 
-fetch_dataset(
+fname = fetch_dataset(
     {"dataset_name": "LFP_data",
      "url": url,
      "archive_name": archive_name,
@@ -34,8 +34,7 @@ fetch_dataset(
     force_update=False
 )
 
-fname = os.path.join(get_config("MNE_DATA"),
-                     "LFP", "LFP_data_contiguous_1250_Hz.mat")
+fname = os.path.join(fname, archive_name)
 
 ###############################################################################
 # It is a mat file, so we use scipy to load it
--- a/setup.cfg
+++ b/setup.cfg
@@ -49,6 +49,8 @@ doc =
     sphinx_gallery
     pactools
     nibabel
+    pooch
+    tqdm
 
 dev =
     flake8
--- a/.github/workflows/deploy_ghpages.yml
+++ b/.github/workflows/deploy_ghpages.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Generate HTML docs
-        uses: ammaraskar/sphinx-action@master
+        uses: ax3l/sphinx-action@newer-sphinx
         with:
           docs-folder: "doc/"
           pre-build-command: |


hndgzkn:
This fixes the problem with running the examples. 

However there are other problems, I guess, due to changes in mne api. 

If you @tomMoral @agramfort confirm, I would update the examples wrt new mne api.

hndgzkn:
@agramfort Are pooch and tqdm dependencies of mne or alphacsc?

agramfort:
mne recently switched to pooch for fetching datasets and pooch depends on
tqdm
but it's not put in install_requires in setup :'(

hndgzkn:
@agramfort  Instead of adding these dependencies in alphacsc, shall we pin documentation building to a stable version of mne?

agramfort:
pip install mne

at this point only installs numpy and scipy as you don't need more for most
cases

there is no such

pip install mne[doc]

solution as the full mne has a recommended install via conda. I will bring
up this issue
on the mne tracker.

cc @larsoner

>

hndgzkn:
@agramfort I've accidentially rebased the branch without getting your changes and force pushed. 

Until mne has a full installation option, how shall we proceed? 
Shall I add dependencies and fix api changes or  pin to an earlier version of mne?

agramfort:
>
> Shall I add dependencies and fix api changes or pin to an earlier version
> of mne?
>

I would add dependencies and fix api changes

>

agramfort:
circleci complains? any idea @hndgzkn

hndgzkn:
Yes, it cannot import `_fetch_file` from `mne.utils` in `examples/csc/plot_lfp_data.py`.

I was looking for what to replace it with. I guess with `mne.datasets.fetch_dataset`. 

> circleci complains? any idea @hndgzkn

hndgzkn:
@agramfort Actually, I was trying to understand why mne version on my local is has version `0.23.4` and why it installs `0.24.dev0` on CI. Then I realised that we have pinned the version of mne to `main` branch in `deploy_ghpages.yml` file. 

I do not remember why this is the case, my be it was the case in previous CI file and I did not change it.  Would you like it to stay that way, or remove the pin to main repo?

hndgzkn:
@agramfort OK, I see, we have pinned to main branch to address #28 as mentioned in #37.

agramfort:
I would suggest to use main branch

mne should be release soon and before alphacsc so it's the best short term
solution

thx

>

hndgzkn:
@agramfort I am trying to download data using [fetch_dataset](https://github.com/mne-tools/mne-python/blob/fb98000c6b9998b3a05ab3d887d0e65c4d3d72a0/mne/datasets/_fetch.py#L25) from mne

```
fetch_dataset(
    {"dataset_name": "LFP_data",
     "url": "https://github.com/hitziger/AWL/raw/master/Experiments/data/LFP_data_contiguous_1250_Hz.mat",
     "archive_name": "LFP_data_contiguous_1250_Hz.mat",
     "config_key": "MNE_DATA",
     "folder_name": "LFP",
     "hash": None
    },
    path=None,
    force_update=True
)
```

The output is 

```
Downloading file 'LFP_data_contiguous_1250_Hz.mat' from 'https://github.com/hitziger/AWL/raw/master/Experiments/data/LFP_data_contiguous_1250_Hz.mat' to '/home/hande/mne_data'.
100%|█████████████████████████████████████| 14.4M/14.4M [00:00<00:00, 17.2GB/s]
```

But the file does not exist under `/home/hande/mne_data`. 

Am I using it wrong or is it possible that there is a bug?

agramfort:
you found a bug in mne main branch... I'll tell you when it's fixed. Thanks
for reporting this issue

>

drammock:
@hndgzkn mne-tools/mne-python#9802 has now been merged, so  `fetch_dataset()` should now work as expected (though note that you need not pass `config_key="MNE_DATA"`; it is redundant with the default behavior)

hndgzkn:
@drammock Thanks! I see your fix, I'll update it.

> @hndgzkn [mne-tools/mne-python#9802](https://github.com/mne-tools/mne-python/pull/9802) has now been merged, so `fetch_dataset()` should now work as expected (though note that you need not pass `config_key="MNE_DATA"`; it is redundant with the default behavior)

hndgzkn:
@drammock I made a bug report https://github.com/mne-tools/mne-python/issues/9806 . When I redownload data with `force_update=False`, it raises error.

setup.cfg
```diff
@@ -49,6 +49,8 @@ doc =
     sphinx_gallery
     pactools
     nibabel
+    pooch
+    tqdm
```
tomMoral, line 53: I think installing `mne` from master should already install this dependency no?
Seeing here: https://github.com/mne-tools/mne-python/blob/main/requirements.txt

tomMoral (approved):
LGTM, just a quick question on the dependencies?

setup.cfg
agramfort: no it doesn't not. It's considered an optional dep of mne that is only required for fetching data. I know it's not perfect but...

agramfort (approved):

setup.cfg
hndgzkn: No, I think it still does not install. See https://github.com/mne-tools/mne-python/blob/61608aa4038a888d1c1f7c9dd5bb1a02203719de/setup.py#L53

tomMoral: The way `mne` handle dependencies seems to make no sense but ok, I see why I did not understood :sweat_smile:

```
    hard_dependencies = ('numpy', 'scipy')
    install_requires = list()
    with open('requirements.txt', 'r') as fid:
        for line in fid:
            req = line.strip()
            for hard_dep in hard_dependencies:
                if req.startswith(hard_dep):
                    install_requires.append(req)
```
Let's merge once it is green then !

examples/csc/plot_lfp_data.py
```diff
@@ -14,13 +14,28 @@
...
+fetch_dataset(
+    {"dataset_name": "LFP_data",
+     "url": url,
+     "archive_name": archive_name,
+     "folder_name": folder_name,
+     "hash": None
+     },
+    path=None,
+    force_update=False
+)
```
tomMoral, line 35: `fetch_dataset` returns the path of the data so you could directly use this instead of `get_config` no?
https://github.com/mne-tools/mne-python/blob/main/mne/datasets/_fetch.py#L89

hndgzkn: Yes, you're right; though it returns the directory path rather than file path. I do not know if it is a bug on mne side.

I've changed it.

tomMoral: This one is not a bug :) Dataset can have multiple files so it makes sense to have the folder containing all the files returned.
Thx for fixing this!

tomMoral:
Merging! Thx @hndgzkn  :)

PR #46, opened 2021-10-01
Dicodile backend

rprimet:


hndgzkn:
@tomMoral @agramfort  We have a few questions regarding the pull request:

- Should the behavior of dicodile and alphacsc encoders be consistent, for the implemented functions, in terms errors of returned from API calls? For example, for alphacsc encoder, as `z_hat` is initialized when the encoder is created, it is possible to call `get_z_hat` before calling `compute_z`. Whereas for dicodile, `z_hat` is initialized when `compute_z` is called. A call to `get_z_hat` or `get_cost` before calling `compute_z` gives error. Shall we keep dicodile behavior this way or initialize `z_hat` before calling `compute_z`? 

- Which of the parameters in `get_encoder_for` and `learn_d_z_multi` are not relevant for dicodile encoder?

tomMoral (commented):
I guess the main difference between the two examples is the initialization.

By default, `learn_d_z_multi` is initializing `D_hat` randomly (https://github.com/alphacsc/alphacsc/blob/master/alphacsc/init_dict.py#L72, `D_init=None`) while the example in `dicodile` uses chunks by default (see https://github.com/tomMoral/dicodile/blob/main/examples/plot_gait.py#L74, `init_dictionary` takes chunks by default). Could you try this?

If it still does not converge, could you paste an image of the obtained dictionary?

examples/dicodile/plot_gait.py
```diff
@@ -0,0 +1,97 @@
...
+# set individual atom (patch) size
+n_times_atom = 200
+
+""
+
+solver_z_kwargs = {}
+
+solver_z_kwargs["max_iter"] = 100000
+
+pobj, times, D_hat, z_hat, reg = learn_d_z_multi(X,
```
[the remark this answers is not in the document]
hndgzkn: Ok. I fixed it.

hndgzkn:
Ok. It works. 

However instead of specifiying `D_init="chunk"` while instantiating CDL, I created an initial dictionary, to make the example similar to  https://tommoral.github.io/dicodile/auto_examples/plot_gait.html#signal-reconstruction-multichannel to view the two dictionaries D_init and D_hat. Let me know if it is unnecessary, I'll remove it.


> If it still does not converge, could you paste an image of the obtained dictionary?

tomMoral:
No I think that this is great! :)

could you share the resulting plot here? If this is working well, I am in favor of merging this (long) PR!!!

alphacsc/convolutional_dictionary_learning.py
```diff
@@ -341,20 +341,23 @@ class BatchCDL(ConvolutionalDictionaryLearning):
     __doc__ = DOC_FMT.format(**_default)
 
     def __init__(self, n_atoms, n_times_atom, reg=0.1, n_iter=60, n_jobs=1,
+                 loss='l2', loss_params=dict(gamma=.1, sakoe_chiba_band=10,
+                                             ordar=10),
```
tomMoral, line 345: I would actually remove these parameters as the code for `loss != 'l2'` is not really documented.

tomMoral (approved):
Beside this one comment, not sure if there is anymore things to do in this first PR?

alphacsc/convolutional_dictionary_learning.py
hndgzkn: As we check the parameters when the solver is `dicodlle` I've put this in order to be able to set loss_params=None. Otherwise we should set loss_params=None everywhere.

hndgzkn:
This is the image. I also did the multichannel part, but there is a problem with activations. See the second image.

signal reconstruction

multi

alphacsc/convolutional_dictionary_learning.py
tomMoral: Ok I get it! I would say that having to set `loss_params=None` in every call seems sub-optimal as there is a high probability it is never used.

So if we could have a default to `None` and set the default value to `dict(gamma=.1, sakoe_chiba_band=10, ordar=10)` if `loss != 'l2'` at the beginning of `learn_d_z_multi`, I guess this would make it easier? WDYT?

hndgzkn:
> why do you think there is an issue with the activation?

- It is different than the image [here](https://tommoral.github.io/dicodile/auto_examples/plot_gait.html#signal-reconstruction-multichannel).
- I thought the activations should be closely aligned with the peaks.

alphacsc/convolutional_dictionary_learning.py
hndgzkn: Ok!

examples/dicodile/plot_gait.py
```diff
@@ -123,11 +122,10 @@
 ###############################################################################
 # Nor, let's reconstruct the original signal.
```
tomMoral, line 123:
```suggestion
# Now, let's reconstruct the original signal.
```

tomMoral:
For the example, could you report the run time for alphacsc and dicodile?

hndgzkn:
> For the example, could you report the run time for alphacsc and dicodile?

Something like this? The above output belongs to `dicodile` solver
times

tomMoral:
yep that was a dirty hack I guess...

I would say put `event_info = dict(event=event, event_id=event_id)` and then `info['tmp'] = event_id` maybe?

or simply propagate correctly the `event_info`? that ask for a bit more plumbering but it will be cleaner in the long run.

agramfort (approved):
perfect fix for new mne API !

thx @hndgzkn  and @rprimet !

agramfort:
is it expected to still have one CI running?

if all green feel free to click the green button

🙏

PR #47, opened 2021-10-04
 build: use PEP517/518 conventions

deepcharles:
Hello everyone,

In this PR, I propose to slightly change the build process to follow [PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/).

### `setup.cfg` and `pyproject.toml`
The now standartized way to package library is using `setup.cfg`, `pyproject.toml` (and optionally `setup.py`). Basically, all that is needed to build the package is installed in a separate virtualenv, the package is then build (in a wheel format) and the virtualenv is removed. Build dependencies (such as cython) do not remain in the user's workspace any more.

To that end, I moved most of `setup.py`'s content into `setup.cfg`. Infortunately, the cython extensions could not be adapted, so they remained in `setup.py`.

To build and install `alphacsc':
```
# for a normal installation
python -m pip install . 
# for an editable installation
python -m pip install -e .
# to also install the test librairies
python -m pip install -e .[test]
```
(Note that the `python -m` can be omitted most of the times.)
The old way of building and installing (`python setup.py install`) will not work any more.

### Version number

I removed the function that parsed the version number (`find_version()` in `setup.py`) as this can be done automatically with the following line in `setup.cfg`:

```yaml
version = attr: alphacsc.__version__
```

### Workflows

It is not needed to separately install cython and numpy beforehand, therefore all lines `pip install numpy cython` are removed from worflows and documentation.

### DIFF
--- a/.github/workflows/deploy_ghpages.yml
+++ b/.github/workflows/deploy_ghpages.yml
@@ -18,7 +18,6 @@ jobs:
           pre-build-command: |
             apt-get update
             apt-get install -y gcc git
-            pip install numpy cython
             pip install -e .[doc]
             pip install git+https://github.com/mne-tools/mne-python.git@main
       - name:  Upload generated HTML as artifact
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -24,7 +24,6 @@ jobs:
         run: |
           conda --version
           which python
-          pip install numpy cython
           pip install -e .[test]
       - name: Run unit tests
         run: |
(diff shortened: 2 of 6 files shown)


deepcharles:
Fell free to tell me what I can change to improve this PR.

agramfort (approved):
looks good !

let's see if CIs are happy

@hndgzkn what do you think?

just a remark @deepcharles when you open a PR don't do it from your master branch

thx again !

tomMoral (approved):
LGTM! Thx @deepcharles

tomMoral:
The error with the doc is being handled in #44.
It requires a fix from `mne`. Maybe we can move on with this PR now as this won't change the state in master but it improves the packaging?

PR #49, opened 2021-10-22
Update compute_d _func

hndgzkn:
Updates `compute_d_func` to give z_encoder instead of X, z_hat, D_hat, constants.

- I would propose to move `compute_d_func` inside z_encoder as `compute_d_update` and not change the signature of `update_uv` and `update_d`.
- As far as I remember, we had also discussed moving `get_lambda_max` and `get_max_error_dict` in encoder API.

If these are OK with you, I'll continue. However, I wonder if you would have time soon to review and merge https://github.com/alphacsc/alphacsc/pull/46. If yes, I would prefer to wait until you merge to avoid too many merge conflicts.

### DIFF
--- a/alphacsc/_encoder.py
+++ b/alphacsc/_encoder.py
@@ -180,7 +180,7 @@ class BaseZEncoder:
         self.n_atoms = n_atoms
         self.atom_support = atom_support
         self.n_jobs = n_jobs
-        self.z_alg = solver
+        self.solver = solver
 
         self.solver_kwargs = solver_kwargs
         self.algorithm = algorithm
@@ -370,7 +370,7 @@ class AlphaCSCEncoder(BaseZEncoder):
             self.D_hat,
             reg=reg,
             z0=z0,
-            solver=self.z_alg,
+            solver=self.solver,
             solver_kwargs=self.solver_kwargs,
             freeze_support=unbiased_z_hat,
             loss=self.loss,
--- a/alphacsc/_encoder.py
+++ b/alphacsc/_encoder.py
@@ -183,9 +183,7 @@ class BaseZEncoder:
         self.n_trials, self.n_channels, self.n_times = X.shape
         self.n_times_valid = self.n_times - self.atom_support + 1
 
-        self.constants = {}
-        self.constants['n_channels'] = self.n_channels
-        self.constants['XtX'] = np.dot(X.ravel(), X.ravel())
+        self.XtX = np.dot(X.ravel(), X.ravel())
 
     def compute_z(self):
         """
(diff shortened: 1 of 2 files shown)


tomMoral (commented):
This looks pretty good! I think this makes a clean pipe.

Now we would need to update `update_uv` and `update_d` to not require `z_hat` when it is not necessary. Should we do this in this PR or in follow up one?

tomMoral:
Oh the failure in the doc seems legitimate, do you reproduce it locally?

hndgzkn:
> Oh the failure in the doc seems legitimate, do you reproduce it locally?

Sorry, somehow I had removed my dev installation, the tests passed on local. 

The problem is due to difference in `compute_z` behavior in alphacsc and dicodile backends. In alphacsc, when we call `compute_z`, `ztz` and `ztX` are set. In dicodile, they are set when you call `get_sufficient_statistics`. It is fixed now.

hndgzkn:
> This looks pretty good! I think this makes a clean pipe.
> 
> Now we would need to update `update_uv` and `update_d` to not require `z_hat` when it is not necessary. Should we do this in this PR or in follow up one?

Ok for doing it in the following PR.

hndgzkn:
@tomMoral Could we merge this if the changes are ok?

Deploy GH pages fail due to https://github.com/sphinx-gallery/sphinx-gallery/issues/879.

alphacsc/_encoder.py
```diff
@@ -148,13 +148,43 @@ def get_z_encoder_for(
...
+        self.n_atoms = n_atoms
+        self.atom_support = atom_support
+        self.n_jobs = n_jobs
+
+        self.solver_kwargs = solver_kwargs
+        self.algorithm = algorithm
+        self.reg = reg
+        self.loss = loss
+
+        self.n_trials, self.n_channels, self.n_times = X.shape
```
agramfort, line 183: I would not store self.n_trials, self.n_channels, self.n_times as it's available in self.X
to make sure the self.X and self.n_trials, self.n_channels, self.n_times are always in sync
I would consider properties but for now just do:

n_trials, n_channels, n_times = self.X

when you need it

agramfort (approved):
just a nitpick

I like red lines :)

thx @hndgzkn

tomMoral (approved):
LGTM! thx @hndgzkn

tomMoral:
Merging, let's fix the nitpick in the next PR. with updated `update_uv` and `update_d`.

tomMoral:
another nitpick: could we try to avoid having to pass `algorithm` to the encoder? I guess if we give pass both `n_atoms` and `D_hat` at initialization, we could have a `D_hat` with shape `(0, n_channels, n_times_atom)` and get a simplified `_init_z_hat`?

PR #55, opened 2021-10-28
MNT adding a procedure to release to PyPI

deepcharles:
The procedure to upload a new version to PyPI is described in `RELEASE.md`.

The version number is computed from the latest tag (see the [versioning scheme here](https://github.com/pypa/setuptools_scm/#default-versioning-scheme])) using `setuptools_scm`. 

In the Github action, the wheels are build for all major platforms and Python versions.
There is an additional job for `aarch64` because it needs an emulator.

The wheels and dist are first uploaded to Test PyPI then to PyPI. In order to test this action, the last job (`Publish distribution to PyPI`) should be commented out first.


.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,110 @@
...
+  build_aarch64_wheels:
+    name: Build wheels manylinux_aarch64
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [36, 37, 38, 39]
+        include:
+          - os: ubuntu-latest
+            arch: aarch64
+            platform_id: manylinux_aarch64
```
tomMoral, line 71: what is the difference with the previous wheels in here?

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,110 @@
...
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+        # do not fail if the TestPyPi release already
+        # exists
+        skip_existing: true
+    - name: Publish distribution to PyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        user: __token__
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
tomMoral, line 110: Not sure why we push to both test pypi and pypi here?
Also what is the token in `user`?

RELEASE.md
```diff
@@ -0,0 +1,14 @@
+# Release procedure
+
+This page describes the procedure to release to PyPI.
+
+### Automatic procedure
+
+Create a [Github release](https://github.com/alphacsc/alphacsc/releases) with a version tag (e.g. "v1.2.3").
```
tomMoral, line 7: should the tag include a `v` and then the version or can we just tag `1.2.3`? If so, what will be the final version  (`v1.2.3` or `1.2.3`)

.gitignore
```diff
@@ -22,3 +22,4 @@ joblib
 __pycache__
 .pytest_cache
 .ipynb_checkpoints
+alphacsc/_version.py
```
tomMoral, line 25:
```suggestion
alphacsc/_version.py

```

tomMoral (approved):
LGTM!

One nitpick and a few curiosity questions in comments.

.github/workflows/upload-to-pypi.yml
deepcharles: The push to test pypi is merely a check that everything goes well before pushing to pypi. It can be removed.

>  Also what is the token in user?

This is the way this action is used in the documentation (see [here](https://github.com/pypa/gh-action-pypi-publish#usage)). It relies on the new API of pypi to authenticate (https://pypi.org/help/#apitoken).

RELEASE.md
deepcharles: Actually, it does not make any difference. whether the tag is `v1.2.3` or `1.2.3`, the version will be `1.2.3`. I usually prefix version tags with a `v`, but it can be dropped without consequences.

.github/workflows/upload-to-pypi.yml
deepcharles: This adds wheels for `manylinux2014_aarch64` platforms. I believe this adds supports for [ARMv8 architectures](https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy).

.github/workflows/upload-to-pypi.yml
tomMoral: Ok I get it for the `__token__`, this is merely to notify that the authentification does not require a user.

I think I would prefer that we don't push to test.pypi, it feels not really useful.

RELEASE.md
tomMoral: Ok thx!

.github/workflows/upload-to-pypi.yml
```diff
@@ -0,0 +1,110 @@
...
+        name: artifact
+        path: dist
+    - name: Publish distribution to Test PyPI
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        password: ${{ secrets.TEST_PYPI_API_TOKEN }}
+        repository_url: https://test.pypi.org/legacy/
+        # do not fail if the TestPyPi release already
+        # exists
+        skip_existing: true
```
tomMoral, line 105:
```suggestion
```

PR #56, opened 2021-11-13
CLN remove lil switch+improve encoders

tomMoral:
I did a pass on cleaning up some of the unused code that made the code more complex (in particular `use_sparse=True`).
I ended up changing a bunch a different stuff to make the code a cleaner:

- remove `use_sparse_z` and and simplify the code. Now, one should check that there is no leftover calls to `lil` matrices and remove the unused code?
- Remove `algorithm` from the encoder. Semantically, itshould not matter for the encoder.
- Add a `get_z_nnz` in the encoder to allow for reporting without accessing `z_hat`.
- Move `get_max_error_patch` in the encoder to also avoid accessing `z_hat` for greedy algorithm. This should be added in the `DicodileEncoder` and in `dicodile` as some point, we would need to have each worker compute its `max_error_patch` and return it with the error, and then only return the one with the max error.
- Add a `check_solver_and_constraints` to handle the `solver_d='auto'` and `uv_constraint='auto'` more reliably.
- Use `get_D_shape` everywhere to get the shape of `D` in rank-1` and non rank-1 cases.
- Change `atom_support` to `n_times_atom` in encoder API to be consistent with the `alphacsc` API.

This is mostly small changes a bit everywhere, if you could review this @hndgzkn and @rprimet, this would be nice!

### DIFF
--- a/alphacsc/tests/test_init_d.py
+++ b/alphacsc/tests/test_init_d.py
@@ -6,19 +6,32 @@ from numpy.testing import assert_allclose
 
 from alphacsc.init_dict import init_dictionary
 from alphacsc.update_d_multi import prox_uv, prox_d
+from alphacsc.update_d_multi import check_solver_and_constraints
 from alphacsc.learn_d_z_multi import learn_d_z_multi
 from alphacsc.utils import check_random_state
 
 
-@pytest.mark.parametrize("rank1", [True, False])
-@pytest.mark.parametrize("uv_constraint", ['separate', 'joint'])
-def test_init_array(rank1, uv_constraint):
+@pytest.mark.parametrize(
+    'rank1, solver_d, uv_constraint',
+    [
+        (True, 'auto', 'auto'),
+        (False, 'auto', 'auto'),
+        (False, 'fista', 'auto'),
+        (True, 'joint', 'joint'),
+        (True, 'joint', 'separate'),
+        (True, 'alternate_adaptive', 'separate')
+    ])
+def test_init_array(rank1, solver_d, uv_constraint):
     n_trials, n_channels, n_times = 5, 3, 100
     n_times_atom, n_atoms = 10, 4
 
+    _, uv_constraint_ = check_solver_and_constraints(
+        rank1, solver_d, uv_constraint
+    )
+
     if rank1:
         expected_shape = (n_atoms, n_channels + n_times_atom)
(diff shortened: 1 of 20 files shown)


alphacsc/update_d_multi.py
```diff
@@ -22,6 +19,30 @@ def squeeze_all_except_one(X, axis=0):
     return X.squeeze(axis=squeeze_axis)
 
 
+def check_solver_and_constraints(rank1, solver_d, uv_constraint):
```
rprimet, line 22: nice!

alphacsc/update_d_multi.py
```diff
@@ -22,6 +19,30 @@ def squeeze_all_except_one(X, axis=0):
...
+                assert uv_constraint == 'separate', (
+                    "solver_d='alternate*' should be used with "
+                    f"uv_constraint='separate'. Got '{uv_constraint}'."
+                )
+        elif uv_constraint == 'auto' and solver_d in ['joint', 'fista']:
+            uv_constraint = 'joint'
+    else:
+        assert solver_d in ['auto', 'fista'] and uv_constraint == 'auto', (
+            "If rank1 is False, uv_constraint should be 'auto' "
+            f"and solver_d should be auto or fista. Got {uv_constraint}"
```
rprimet, line 40: maybe output the value of `solver_d` here as well?
```suggestion
            f"and solver_d should be auto or fista. Got {solver_d=}, {uv_constraint=}"
```

rprimet (commented):
Nice!

Just caught a few typos

alphacsc/tests/test_init_d.py
```diff
@@ -6,19 +6,32 @@
...
-def test_init_array(rank1, uv_constraint):
+@pytest.mark.parametrize(
+    'rank1, solver_d, uv_constraint',
+    [
+        (True, 'auto', 'auto'),
+        (False, 'auto', 'auto'),
+        (False, 'fista', 'auto'),
+        (True, 'joint', 'joint'),
+        (True, 'joint', 'separate'),
+        (True, 'alternate_adaptive', 'separate')
```
tomMoral, line 22: Maybe we should move this as a fixture? wdyt @rprimet

tomMoral: done

tomMoral:
Merging as this is green and blocking the other. Feel free to correct any mistakes and clean up unused files (such as `lil.py` or some `cython` code).

PR #57, opened 2021-11-17
ENH DSolver

hndgzkn:
Creates `DSolver` class depending on `solver_d` and moves `update_d` inside `DSolver`.

If the changes are ok with you, I'll continue to document and add some tests for solver d.

Some questions:
- I moved `compute_lipschitz` inside solver_d. I can put it back, or I can move other functions in `update_d_multi.py` if they are not expected to be used outside solver_d. `update_uv` and `update_d` are no more used from `learn_d_z_multi.py`.
- I'm not sure about having `z_encoder` in the `update_D` signature. I'm planning to change it to use `X`, `D_hat`, and instead of `z` `get_z_hat` function, etc.
- `b_hat_0` seems to be only used from alternate solver (`AlternateDSolver`), would it be ok to initialize it inside the constructor of `AlternateDSolver` ? Or is it preferred to be able to set from outside?


tomMoral:
> Creates `DSolver` class depending on `solver_d` and moves `update_d` inside `DSolver`.

I really like the idea of having separate `DSolver` for `rank1` and not rank-1, thanks for the proposition.

>     * I moved `compute_lipschitz` inside solver_d. I can put it back, or I can move other functions in `update_d_multi.py` if they are not expected to be used outside solver_d. `update_uv` and `update_d` are no more used from `learn_d_z_multi.py`.

Seems the right move as it is only used for this solver.

>     * I'm not sure about having `z_encoder` in the `update_D` signature. I'm planning to change it to use `X`, `D_hat`, and instead of `z` `get_z_hat` function, etc.

I am more in favor of giving the `z_encoder` as each solver will need to decide which part it needs (sufficient statistics, constants, ...). I would stick to using the `z_encoder`. But maybe we can pass `X` and `D_hat` separately?

>     * `b_hat_0` seems to be only used from alternate solver (`AlternateDSolver`), would it be ok to initialize it inside the constructor of `AlternateDSolver` ? Or is it preferred to be able to set from outside?

No it is right to move it to `AlternateDSolver`.


I will do a review ASAP.

hndgzkn:
Thanks for the answers. 

Two more questions:

- Are the variables in the constructor good to store as class variables, or it would be ok to move some of them (like `eps`,  `max_iter`,  `momentum`) to `update_D` signature. Might they change for different calls to `update_D`?
- Would it be reasonable to put `init_dictionary` inside `solver_d`?

tomMoral:
> Are the variables in the constructor good to store as class variables, or it would be ok to move some of them (like `eps`,  `max_iter`,  `momentum`) to `update_D` signature. Might they change for different calls to `update_D`?

I would keep them in the constructor for now as they don't move for now.
We can add setters or move them if we need.

>   Would it be reasonable to put `init_dictionary` inside `solver_d`?

I think so yes.

alphacsc/_solver_d.py
```diff
@@ -0,0 +1,692 @@
...
+            maximal reconstruction error.
+
+        [Yellin2017] BLOOD CELL DETECTION AND COUNTING IN HOLOGRAPHIC LENS-FREE
+        IMAGING BY CONVOLUTIONAL SPARSE DICTIONARY LEARNING AND CODING.
+        """
+        n_times_atom = z_encoder.n_times_atom
+        d0 = z_encoder.get_max_error_patch()
+
+        if self.window:
+            d0 = d0 * tukey_window(n_times_atom)[None, :]
```
tomMoral, line 632: It would be nice to compute this window once for all and store it in the class no?

alphacsc/_solver_d.py
```diff
@@ -0,0 +1,692 @@
...
+        def prox_v(v, step_size=None):
+            if self.window:
+                v *= tukey_window_
+
+            v /= np.maximum(1., np.linalg.norm(v, axis=1, keepdims=True))
+
+            if self.window:
+                v /= tukey_window_
+
+            return v
```
tomMoral, line 463: Can we move this as class methods?

alphacsc/_solver_d.py
```diff
@@ -0,0 +1,692 @@
...
+            u_hat, pobj_u = fista(obj, grad_u, prox_u, 0.99 / Lu, u_hat,
+                                  self.max_iter, momentum=self.momentum,
+                                  eps=self.eps,
+                                  adaptive_step_size=adaptive_step_size,
+                                  debug=debug, verbose=verbose,
+                                  name="Update u")
+            uv_hat = np.c_[u_hat, v_hat]
+            if debug:
+                pobj.extend(pobj_u)
+
```
tomMoral, line 508: To make this code easier to read, maybe move each update step in separate functions?

alphacsc/update_d_multi.py
```diff
@@ -215,6 +242,8 @@ def grad_u(u):
             if adaptive_step_size:
                 Lu = 1
             else:
+                if loss != 'l2':
+                    raise NotImplementedError()
```
tomMoral, line 246: Haven't all this code been moved in `_solver_d.py`? I would havoid having duplicated code here if this is not to be called anymore.

tomMoral (commented):
A pass of comments on the code.
Overall, this looks very nice! A few question of API and moving around some stuff.

hndgzkn:
For removing `feasible_evaluation` and `uv_constraint` form the `z_encoder`: I'am not sure. `get_cost` function uses them both. I also added `compute_objective` function and it uses as well. 

However I removed it from `solver_d`.

tomMoral (commented):
I did a pass on the PR, this is nice. I like the Window API.
There is a few changes to do:

- `uv_constraint` and `feasible_evaluation` should not be parameters of the `Zencoder`. I detailled why and how to do in the comments.
- partial_fit does not use `compute_z_partial`.
- Some more cleaning in the API and I think a few extra doctring would be nice.

Let me know when you want me to review again.

alphacsc/online_dictionary_learning.py
```diff
@@ -62,52 +63,46 @@ def partial_fit(self, X, y=None):
...
-        return z_hat
+            z_encoder.compute_z()
+
+            alpha = self.algorithm_params['alpha']
+            self.constants['ztz'] = alpha * \
+                self.constants['ztz'] + z_encoder.ztz
+            z_encoder.ztz = self.constants['ztz']
+            self.constants['ztX'] = alpha * \
+                self.constants['ztX'] + z_encoder.ztX
+            z_encoder.ztX = self.constants['ztX']
```
tomMoral, line 79: I think it would be more efficient to only inline operations instead of allocating 3 arrays?
```suggestion
            self.constants['ztz'] *= alpha
            self.constants['ztz'] += z_encoder.ztz
            z_encoder.ztz = self.constants['ztz']
            self.constants['ztX'] *= alpha
            self.constants['ztX'] += z_encoder.ztX
            z_encoder.ztX = self.constants['ztX']
```

Also, I am not sure why we should

alphacsc/tests/test_solver_d.py
```diff
@@ -0,0 +1,269 @@
...
+        get_solver_d(solver_d='auto',
+                     uv_constraint=uv_constraint,
+                     rank1=False,
+                     window=True,
+                     momentum=False)
+
+
+@pytest.mark.parametrize('solver_d', ['alternate', 'alternate_adaptive',
+                                      'joint'])
+def test_get_solver_d_error_rank1_uv_constraint(solver_d):
```
tomMoral, line 72:
```suggestion
def test_get_solver_d_error_rank1_uv_constraint():
```

alphacsc/tests/test_solver_d.py
```diff
@@ -0,0 +1,269 @@
...
+                     momentum=False)
+
+
+@pytest.mark.parametrize('uv_constraint', ['separate', 'joint'])
+def test_get_solver_d_error_uv_constraint(uv_constraint):
+    """Tests for the case rank1 is False and uv_constraint is not
+    compatible."""
+
+    with pytest.raises(AssertionError,
+                       match="solver_d should be auto or fista. Got*"):
```
tomMoral, line 61: It does not seem to be the right error message no? Should be something about bad `uv_constraint`?

alphacsc/tests/test_online_dictionary_learning.py
```diff
@@ -1,28 +1,24 @@
...
 import numpy as np
 
-from alphacsc.utils import check_random_state
 from alphacsc.online_dictionary_learning import OnlineCDL
+from alphacsc.tests.conftest import N_TIMES_ATOM, N_ATOMS
 
 
-@pytest.mark.parametrize('rank1', [True, False])
+@pytest.mark.parametrize('solver, rank1, n_trials', [
+    ('lgcd', True, 10),
```
tomMoral, line 9:
```suggestion
    ('lgcd', True, 10),
    ('lgcd', False, 10),
```

tomMoral (approved):
mostly good to me. My last comments are:

- `rank1` should only be passed when needed in `DSolver`. When it is fixed, we should avoid passing it.
- I would not pass `uv_constraint` to `get_cost` and ensure we use `feasbile_evaluation=False` if necessary, as this function is really to compute the loss of the encoder with its current dictionary which should already be feasible. I would also not pass it in `ZEncoder.compute_objective` as it would bring a discrepancy between the `l2` loss which does not ensure the feasability and the other ones.

alphacsc/_encoder.py
```diff
@@ -485,6 +501,9 @@ def __init__(
                          reg,
                          loss)
 
+        self.uv_constraint = 'auto'
+        self.loss_params = None
```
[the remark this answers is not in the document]
hndgzkn: I tried what we talked, and there are some problems, some tests did not pass and I could not quickly understand why. I think we can work on that in another PR, but not this one.

alphacsc/online_dictionary_learning.py
```diff
@@ -54,60 +54,58 @@ def __init__(self, n_atoms, n_times_atom, reg=0.1, n_iter=60, n_jobs=1,
             lmbd_max=lmbd_max, raise_on_increase=False, loss='l2',
             callback=None, verbose=verbose, name="OnlineCDL"
         )
+        self.index = 0
 
-    def partial_fit(self, X, y=None):
+    def partial_fit(self, X, i0, y=None):
```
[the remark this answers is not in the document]
hndgzkn: I reverted

alphacsc/online_dictionary_learning.py
```diff
@@ -54,60 +54,58 @@ def __init__(self, n_atoms, n_times_atom, reg=0.1, n_iter=60, n_jobs=1,
...
+                               feasible_evaluation=False) as z_encoder:
+
+            if hasattr(self, 'constants'):
+                z_encoder.ztz = self.constants['ztz']
+                z_encoder.ztX = self.constants['ztX']
+                if hasattr(self, 'z_hat'):
+                    z_encoder.z_hat = self.z_hat
+
+            z_encoder.compute_z_partial(
+                [i0], alpha=self.algorithm_params['alpha'])
```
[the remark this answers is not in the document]
hndgzkn: Fixed

alphacsc/_solver_d.py
```diff
@@ -0,0 +1,635 @@
...
+        super().__init__(solver_d,
+                         rank1,
+                         uv_constraint,
+                         window,
+                         eps,
+                         max_iter,
+                         momentum,
+                         random_state,
+                         verbose,
+                         debug)
```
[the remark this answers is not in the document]
hndgzkn: In another pull request, I might put contents of `init_dictionary` in solver_d and initialize dictionary depending on solver. That way I can remove the `rank1` from class attributes completely.

alphacsc/online_dictionary_learning.py
```diff
@@ -121,6 +116,8 @@ def _ensure_fit_init(self, X):
         self.n_channels_ = n_channels = X.shape[1]
 
         self.constants['n_channels'] = n_channels
+        self.constants['XtX'] = X.ravel().dot(X.ravel())
```
tomMoral, line 119: Actually, the constant `XtX` should be updated as done for the other constants so:
```suggestion
        self.constants['XtX'] = 0
```

alphacsc/online_dictionary_learning.py
```diff
@@ -62,52 +63,46 @@ def partial_fit(self, X, y=None):
...
-        return z_hat
+            z_encoder.compute_z()
+
+            alpha = self.algorithm_params['alpha']
+            self.constants['ztz'] *= alpha
+            self.constants['ztz'] += z_encoder.ztz
+            z_encoder.ztz = self.constants['ztz']
+            self.constants['ztX'] *= alpha
+            self.constants['ztX'] += z_encoder.ztX
+            z_encoder.ztX = self.constants['ztX']
```
tomMoral, line 79:
```suggestion
            z_encoder.ztX = self.constants['ztX']
            self.constants['XtX'] *= alpha
            self.constants['XtX'] += z_encoder.XtX
            z_encoder.XtX = self.constants['XtX']
```

tomMoral (approved):
A few more comments/nitpicks but other wise LGTM!

For formatting, I don't really like the column of argument. I find this too verbose.
Do you have a strong opinion on this?

alphacsc/_encoder.py
```diff
@@ -95,63 +82,48 @@ def get_z_encoder_for(
...
-        )
+        return DicodileEncoder(X,
+                               D_hat,
+                               n_atoms,
+                               n_times_atom,
+                               n_jobs,
+                               solver_kwargs,
+                               reg,
+                               loss,
+                               loss_params)
```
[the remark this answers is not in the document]
hndgzkn: I find it easier to read. However if you prefer otherwise, I'll change it and other similar occurences.

alphacsc/_encoder.py
```diff
@@ -385,16 +380,13 @@ def compute_z_partial(self, i0, alpha=.8):
...
-                                             feasible_evaluation=True,
-                                             return_X_hat=False)
-        return cost
+
+        X_hat = construct_X_multi(self.z_hat, D=self.D_hat,
+                                  n_channels=self.n_channels)
+
+        return compute_objective(X=self.X, X_hat=X_hat, z_hat=self.z_hat,
+                                 reg=self.reg, loss=self.loss,
+                                 loss_params=self.loss_params)
```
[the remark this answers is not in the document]
hndgzkn: `ztz` and `ztx` are not initialized before calling `compute_z`. 

Shall I initialize them with zeros or raise an exception as in `get_sufficient_statistics` telling that `compute_z` should be called before calling `get_cost` function?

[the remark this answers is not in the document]
tomMoral: Yes that would work.

tomMoral (approved):
LGTM! merging, thx @hndgzkn !

PR #58, opened 2021-11-28
ENH Ensure v_hat_ peaks in positive

CedricAllain:
Update v_hat_(self) to ensure that, for each atom k, the corresponding temporal pattern v_hat_[k] peaks in the positive. If not the case, v_hat_[k] is flip by multiplying by -1.

### DIFF
--- a/alphacsc/convolutional_dictionary_learning.py
+++ b/alphacsc/convolutional_dictionary_learning.py
@@ -4,6 +4,7 @@
 #          Alexandre Gramfort <alexandre.gramfort@inria.fr>
 #          Thomas Moreau <thomas.moreau@inria.fr>
 
+import numpy as np
 from sklearn.base import TransformerMixin
 from sklearn.exceptions import NotFittedError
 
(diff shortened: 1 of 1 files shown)


alphacsc/convolutional_dictionary_learning.py
```diff
@@ -311,7 +312,16 @@ def v_hat_(self):
...
         """
-        return self.uv_hat_[:, self.n_channels_:]
+        v_hat_ = self.uv_hat_[:, self.n_channels_:]
+        # Ensure that peaks in v_hat_ are positive
+        index_array = np.argmax(np.absolute(v_hat_), axis=1)
+        val_index = np.take_along_axis(v_hat_, np.expand_dims(
+            index_array, axis=-1), axis=-1).squeeze(axis=-1)
+        # Get atoms indices that need to have their v_hat_ flipped
+        index_to_flip = np.where(val_index < 0)[0]
+        v_hat_[index_to_flip] *= -1
```
agramfort, line 322: I would not do this here and especially not in a property that is not expected to change the object like done here.

I would do it when self.uv_hat_ is created.

PR #60, opened 2021-12-16
Dicodile rank1 example

rprimet:
Draft PR for a rank1 example that uses DiCoDiLe

Misc. notes:

- needs an install of the `rank1_support` branch of dicodile (see https://github.com/tomMoral/dicodile/pull/48 )
- fit works, but there is an outstanding issue with transform
- in DicodileEncoder, we store D_hat using the alphaCSC convention (n_atoms, n_channels+n_times_atom)


alphacsc/_encoder.py
```diff
@@ -122,9 +122,9 @@ def get_z_encoder_for(
         assert feasible_evaluation is False, (
             "DiCoDiLe requires feasible_evaluation=False."
         )
-        assert uv_constraint == 'auto',  (
-            "DiCoDiLe requires uv_constraint=auto."
-        )
+        # assert uv_constraint == 'auto',  (
+        #     "DiCoDiLe requires uv_constraint=auto."
+        # ) #XXX
```
agramfort, line 127: shall we remove this?

agramfort (commented):
also I pushed a commit here @rprimet 

how long does the example take to run on your machine?

alphacsc/_encoder.py
rprimet: Sure!

Regarding running time, the fit time was 2670.7s on my laptop (it seems to be the lion's share of the total running time)

Timing it on my machine gave

```
real	48m3,636s
```

agramfort: that's quite long to be honest. dicodile is supposed to be faster that alphacsc. Maybe it's due to greedy vs non greedy.

in any case this can be improved later as the results already look good !

rprimet: Yes, it would be interesting to see how this affects running time (5-way split in the original example vs no split here). I did not yet time the original example on my laptop but on the CI machine it's around 70 minutes I think?

agramfort:
to be timed next

rprimet:
Notes for investigating performance (from @tomMoral)

> a few leads:
>
> - Could you log the time per step (z and d step) so we can see if we spend more time on D or z?
> max_iter=100 is too low for dicodile, this means1 iterations per worker for n_jobs = 128. I guess what happens is that as we > are not precise enough for the z-step, the algorithm do not converge so we always take the same time.
> - Having a plot with the loss after each step would help a lot to see what is happening

tomMoral (approved):
A few comments but good to go for me, we can improve the example afterward.

alphacsc/_z_encoder.py
```diff
@@ -444,7 +444,17 @@ def __init__(self, X, D_hat, n_atoms, n_times_atom, n_jobs,
...
         self.params = params
-        self._encoder.init_workers(self.X[0], self.D_hat, reg, self.params)
+        self._encoder.init_workers(self.X[0], self._as_dicodile_dict(),
+                                   reg, self.params)
+
+    def _as_dicodile_dict(self):
+        n_channels = self.X.shape[1]
+        if self.D_hat.ndim == 2:  # AlphaCSC convention for rank-1 matrices
+            # Dicodile convention: rank-1 dicts are tuples
+            return (self.D_hat[:, :n_channels],  # XXX ?
```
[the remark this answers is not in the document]
rprimet: That was just me not being convinced of my numpy chops :-) I removed the `XXX`

agramfort:
@rprimet can you share here the figures obtained with this example ?

also can you see why CIs complain ?

🙏

rprimet:
Some runs on margaret (fit time only), with oversubscription

Number of workers | Fit time (s) (marg032 for 32;64;128 cores / marg036 for 8;24 cores)

8   | 1104.0

16  | 1170.2

24  | 1029.3

32  |  885.9

64  |  886.6

128 | 1601.1
visualization

rprimet:
`build_docs` error may be possibly related to sphinx? (see https://github.com/librosa/librosa/issues/1490)
I'll attempt a local build with sphinx <1.5

tomMoral (approved):
Let's merge this one and investigate the performance issues in the benchmarks

PR #62, opened 2022-01-12
ENH change dictionary initialization

hndgzkn:
Changes dictionary initialization using `DictGenerator` and `Rank1DictGenerator` that exploits strategies depending on the value of `D_init`.

I've first moved `init_dictionary` inside `d_solver`, however it seems like dictionary initialization might be necessary outside of `d_solver` as well. So I created generators that can be set upon initialization in `d_solver` depending on `rank1` status. 

I've put window and prox inside `DictGenerator`.  

This PR also binds `d_solver` to  `n_channels`, `n_atoms`, `n_times_atom`.

In `BaseDictGenerator` `wrap` and `wrap_rank1` should be named appropriately.

### DIFF
--- a/alphacsc/_d_solver.py
+++ b/alphacsc/_d_solver.py
@@ -1,11 +1,12 @@
 import numpy as np
 
-from .init_dict import InitStrategy, Rank1InitStrategy, get_generator
+from .init_dict import DictStrategy, Rank1DictStrategy
 from .loss_and_gradient import gradient_uv, gradient_d
 from .update_d_multi import prox_d, prox_uv, check_solver_and_constraints
 from .utils import check_random_state
 from .utils.convolution import numpy_convolve_uv
-from .utils.dictionary import NoWindow, UVWindower, SimpleWindower, get_uv
+from .utils.dictionary import get_uv, NoWindow, UVWindower, SimpleWindower
+
 from .utils.optim import fista, power_iteration
 
 
@@ -93,7 +94,6 @@ class BaseDSolver:
         self.n_times_atom = n_times_atom
         self.rank1 = rank1
         self.uv_constraint = uv_constraint
-        self.window = window
         self.eps = eps
         self.max_iter = max_iter
         self.momentum = momentum
@@ -101,7 +101,7 @@ class BaseDSolver:
         self.verbose = verbose
         self.debug = debug
 
-        if not self.window:
+        if not window:
             self.windower = NoWindow()
         else:
             self._init_windower()
@@ -167,8 +167,8 @@ class BaseDSolver:
             The initial atoms to learn from the data.
         """
 
-        dict_generator = get_generator(self.dict_strategy, D_init)
-        D_hat = dict_generator.get_dict(X, D_init_params)
+        self.dict_strategy.set_dict_generator(D_init)
+        D_hat = self.dict_strategy.get_dict(X, D_init_params)
 
         if not isinstance(D_init, np.ndarray):
             D_hat = self.windower.window(D_hat)
@@ -220,7 +220,7 @@ class Rank1DSolver(BaseDSolver):
             eps, max_iter, momentum, random_state, verbose, debug, rank1=True
         )
 
-        self.dict_strategy = Rank1InitStrategy(
+        self.dict_strategy = Rank1DictStrategy(
             self.n_channels, self.n_atoms, self.n_times_atom, self.rng
         )
 
@@ -531,7 +531,7 @@ class DSolver(BaseDSolver):
             eps, max_iter, momentum, random_state, verbose, debug, rank1=False
         )
 
-        self.dict_strategy = InitStrategy(
+        self.dict_strategy = DictStrategy(
             self.n_channels, self.n_atoms, self.n_times_atom, self.rng
         )
 
--- a/alphacsc/_d_solver.py
+++ b/alphacsc/_d_solver.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from .init_dict import Dictionary, Rank1Dictionary
+from .init_dict import DictionaryUtil, Rank1DictionaryUtil
 from .loss_and_gradient import gradient_uv, gradient_d
 from .utils import check_random_state
 from .utils.convolution import numpy_convolve_uv
@@ -166,7 +166,7 @@ class BaseDSolver:
 
         def objective(D, full=False):
 
-            D = self.dictionary.window(D)
+            D = self.dict_util.window(D)
 
             return z_encoder.compute_objective(D)
 
@@ -176,11 +176,11 @@ class BaseDSolver:
 
         def prox(D, step_size=None):
 
-            D = self.dictionary.window(D)
+            D = self.dict_util.window(D)
 
-            D = self.dictionary.prox(D)
+            D = self.dict_util.prox(D)
 
-            return self.dictionary.dewindow(D)
+            return self.dict_util.dewindow(D)
 
         return prox
 
@@ -188,11 +188,11 @@ class BaseDSolver:
 
         def grad(D):
 
-            D = self.dictionary.window(D)
+            D = self.dict_util.window(D)
 
             grad = self.grad(D, z_encoder)
 
-            return self.dictionary.window(grad)
+            return self.dict_util.window(grad)
 
         return grad
 
@@ -220,9 +220,9 @@ class BaseDSolver:
 
         d0 = z_encoder.get_max_error_patch()
 
-        d0 = self.dictionary.window(d0)
+        d0 = self.dict_util.window(d0)
 
-        return self.dictionary.prox(d0)
+        return self.dict_util.prox(d0)
 
     def init_dictionary(self, X):
         """Returns a dictionary for the signal X depending on D_init value.
@@ -239,16 +239,15 @@ class BaseDSolver:
             The initial atoms to learn from the data.
         """
 
-        self.D_hat = self.dictionary.initialize(X)
-
+        self.D_hat = self.dict_util.initialize(X)
         return self.D_hat
 
     def add_one_atom(self, z_encoder):
         new_atom = self.get_max_error_dict(z_encoder)[0]
 
-        self.D_hat = np.concatenate([self.D_hat, new_atom[None]])
-        z_encoder.set_D(self.D_hat)
-        z_encoder.update_z_hat()
+        self.D_hat = self.dict_util.add_one_atom(self.D_hat, new_atom)
+
+        z_encoder.update_z_hat(self.D_hat)
 
     def update_D(self, z_encoder):
         """Learn d's in time domain.
@@ -267,7 +266,7 @@ class BaseDSolver:
 
         assert z_encoder.n_channels == self.n_channels
 
-        D_hat0 = self.dictionary.dewindow(self.D_hat)
+        D_hat0 = self.dict_util.dewindow(self.D_hat)
 
         D_hat, pobj = fista(
             self._get_objective(z_encoder), self._get_grad(z_encoder),
@@ -276,7 +275,7 @@ class BaseDSolver:
             name=self.name, debug=self.debug, verbose=self.verbose
         )
 
-        self.D_hat = self.dictionary.window(D_hat)
+        self.D_hat = self.dict_util.window(D_hat)
 
         z_encoder.set_D(self.D_hat)
 
@@ -297,7 +296,7 @@ class Rank1DSolver(BaseDSolver):
             eps, max_iter, momentum, random_state, verbose, debug
         )
 
-        self.dictionary = Rank1Dictionary(
+        self.dict_util = Rank1DictionaryUtil(
             self.n_channels, self.n_atoms, self.n_times_atom,
             self.rng, window, D_init, D_init_params, self.uv_constraint
         )
@@ -358,7 +357,7 @@ class AlternateDSolver(Rank1DSolver):
         """
         assert z_encoder.n_channels == self.n_channels
 
-        uv_hat = self.dictionary.dewindow(self.D_hat)
+        uv_hat = self.dict_util.dewindow(self.D_hat)
 
         objective = self._get_objective(z_encoder)
 
@@ -377,7 +376,7 @@ class AlternateDSolver(Rank1DSolver):
                 pobj.extend(pobj_u)
                 pobj.extend(pobj_v)
 
-        self.D_hat = self.dictionary.window(uv_hat)
+        self.D_hat = self.dict_util.window(uv_hat)
 
         z_encoder.set_D(self.D_hat)
 
@@ -394,7 +393,7 @@ class AlternateDSolver(Rank1DSolver):
 
         def grad_u(u):
 
-            uv = np.c_[u, self.dictionary.simple_window(v_hat)]
+            uv = np.c_[u, self.dict_util.simple_window(v_hat)]
 
             grad_d = gradient_d(
                 uv, X=z_encoder.X, z=z_encoder.get_z_hat(),
@@ -435,7 +434,7 @@ class AlternateDSolver(Rank1DSolver):
 
         def grad_v(v):
 
-            v = self.dictionary.simple_window(v)
+            v = self.dict_util.simple_window(v)
 
             uv = np.c_[u_hat, v]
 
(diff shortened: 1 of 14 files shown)


alphacsc/init_dict.py
```diff
@@ -24,7 +24,7 @@
 tied = custom_distances.translation_invariant_euclidean_distances
 
 
-class BaseDictGenerator():
+class BaseDictionary():
 
     def __init__(self, n_channels, n_atoms, n_times_atom, random_state,
                  window):
```
tomMoral, line 30: I would pass init parameters here `D_init` and `D_init_params` so we simply pass `X` in the `init` method.

alphacsc/init_dict.py
```diff
@@ -56,24 +56,24 @@ def simple_dewindow(self, D_hat):
...
         raise NotImplementedError()
 
-    def get_D_shape(self):
+    def get_shape(self):
         raise NotImplementedError()
 
-    def get_dict(self, X, D_init_params):
-        D_hat = self.strategy.get_dict(X, D_init_params)
+    def init(self, X, D_init_params):
+        D_hat = self.strategy.init(X, D_init_params)
```
tomMoral, line 63: This is confusing with `__init_`, maybe better to use `initialize`?

alphacsc/init_dict.py
```diff
@@ -132,47 +132,38 @@ def wrap_rank1(self, D_hat):
...
+    def __init__(self, parent):
+        self.parent = parent
+        self.n_channels = parent.n_channels
+        self.n_atoms = parent.n_atoms
+        self.n_times_atom = parent.n_times_atom
+        self.rng = parent.rng
 
-    def get_dict(self, X, D_init_params):
+    def init(self, X, D_init_params):
         raise NotImplementedError()
```
tomMoral, line 143: Not sure it is worth having this extra abstraction?
Maybe better to simply have functions that returns  `D_init` given `cdl_params` containing what is contained in `init`?

tomMoral (commented):
A few comments.

The main one being that I think we should replace `D_hat` by a dictionary object in all non-user facing code. That would simplify the reading no? WDYT?

hndgzkn:
> A few comments.
> 
> The main one being that I think we should replace `D_hat` by a dictionary object in all non-user facing code. That would simplify the reading no? WDYT?

Yes, that was what we talked. I pushed, but I haven't yet finished all changes. I will arrive at that point gradually.

tomMoral:
Ah sorry, too fast review for once :)

alphacsc/init_dict.py
tomMoral: As the `d_solver` is created at the same time as the initialization, I would say it is ok to pass `D_init` and `D_init_params` directly in the constructor.

The rational for this is the following: `D_init` and `D_init_params` are parameters of the `d_solver` so following the `sklearn` API, they should be passed in the constructor. `X` is data related so it should be passed in the `fit` method, which to me correspond to the `initialize_dictionary` method.

alphacsc/_d_solver.py
```diff
@@ -187,19 +187,19 @@ def grad(D):
...
-        return self.dict_generator.simple_window(D_hat)
+        return self.dictionary.simple_window(D_hat)
 
     def simple_dewindow(self, D_hat):
-        return self.dict_generator.simple_dewindow(D_hat)
+        return self.dictionary.simple_dewindow(D_hat)
 
     def prox(self, D):
-        return self.dict_generator.prox(D)
+        return self.dictionary.prox(D)
```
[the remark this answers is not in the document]
hndgzkn: I removed these functions.

hndgzkn:
At the end I preferred not to use dictionary object everywhere in `d_solver`.  

Instead I transformed `dictionary` to `dict_util` which is created depending on dict type being `rank1` or not and with certain shape, and lets you initialize dictionary with a certain strategy, provides some functions for dictionaries (window, remove_window, prox, add_one_atom, get_shape). However it does not hold `D_hat` inside.

I preferred this because actually we use `window`/`remove_window` and `prox` functions not only for `D_hat`, but also for `d0` in `get_max_error_dict` for example, there are some other places. I tried to find some solutions, but the code gets polluted with if conditions very quickly. 

I would like to make some minor changes in `learn_d_z_multi`, but I just wanted to have your opinion on `dict_util` part. After I make changes in `learn_d_z_multi` it will be ready for review.

alphacsc/_d_solver.py
```diff
@@ -1,22 +1,84 @@
 import numpy as np
 
-from .init_dict import init_dictionary
+from .init_dict import get_dict_strategy
```
tomMoral, line 3: I would use `init_strategy` instead of `dict_strategy`.

alphacsc/_d_solver.py
```diff
@@ -183,122 +227,144 @@ def get_max_error_dict(self, z_encoder):
...
+        D : array shape (n_atoms, n_channels + n_times_atom) or
+                  shape (n_atoms, n_channels, n_times_atom)
+            The initial atoms to learn from the data.
+        """
 
-        return objective
+        D_hat = self.dict_strategy.initialize(X)
 
+        if not hasattr(self.dict_strategy, 'D_init'):
+            D_hat = self._windower.window(D_hat)
```
tomMoral, line 256: Not sure what is the rational here, can you add a comment?

alphacsc/_d_solver.py
```diff
@@ -183,122 +227,144 @@ def get_max_error_dict(self, z_encoder):
...
-        )
+    def add_one_atom(self, z_encoder):
+        new_atom = self.get_max_error_dict(z_encoder)[0]
 
-        self.name = "Update uv"
+        self.D_hat = np.concatenate([self.D_hat, new_atom[None]])
 
-    def get_max_error_dict(self, z_encoder):
-        """Get the maximal reconstruction error patch from the data as a new atom
+        z_encoder.update_z_hat(self.D_hat)
```
tomMoral, line 267: I would only use `set_D` and check that the number of component in `z_hat` are enough each time, without special case for `update_z_hat`

alphacsc/_z_encoder.py
```diff
@@ -372,20 +365,25 @@ def get_max_error_patch(self):
...
-    def add_one_atom(self, new_atom):
-        assert new_atom.shape == (self.n_times_atom + self.X.shape[1],)
-        self.D_hat = np.concatenate([self.D_hat, new_atom[None]])
-        self.z_hat = np.concatenate(
-            [self.z_hat, self._get_new_z_hat(1)], axis=1
-        )
+        if self.z_hat.shape[1] < self.n_atoms:
+            self.z_hat = np.concatenate(
+                [self.z_hat, self._get_new_z_hat(1)], axis=1
+            )
```
tomMoral, line 386: I would use a single function `set_D` that checks that `self.z_hat.shape[1] == self.D_hat.shape[1]` and adds columns if needed.

alphacsc/init_dict.py
```diff
@@ -18,10 +18,128 @@
...
+
+        return D_hat
+
+
+class GreedyStrategy(RandomStrategy):
+
+    def initialize(self, X):
+        D_hat = super().initialize(X)
+        return D_hat[:0]
+
```
tomMoral, line 141: This is super neat :)

alphacsc/tests/test_learn_d_z_multi.py
```diff
@@ -13,7 +12,20 @@
...
+                             (False, 'fista', 'auto'),
+                             (True, 'joint', 'auto'),
+                             (True, 'joint', 'joint'),
+                             (True, 'joint', 'separate'),
+                             (True, 'fista', 'auto'),
+                             (True, 'fista', 'joint'),
+                             (True, 'fista', 'separate'),
+                             (True, 'alternate_adaptive', 'separate')
+                         ]
+                         )
```
tomMoral, line 28: Why not use the one defined in `conftest`?

alphacsc/tests/test_learn_d_z_multi.py
```diff
@@ -149,7 +157,8 @@ def test_online_learning():
 
 
 @pytest.mark.parametrize('klass', [BatchCDL, OnlineCDL, GreedyCDL])
-def test_transformers(klass):
+@parametrize_solver_and_constraint
+def test_transformers(klass, rank1, solver_d, uv_constraint):
```
tomMoral, line 161: The extra argument are not passed to the class.

alphacsc/tests/test_update_d_multi.py
```diff
@@ -5,81 +5,13 @@
...
-
-    solver_d_, uv_constraint_ = check_solver_and_constraints(True, solver_d,
-                                                             uv_constraint)
-
-    if uv_constraint == 'auto':
-        uv_constraint = 'joint'
-
-    assert solver_d_ == solver_d
-    assert uv_constraint_ == uv_constraint
-
```
tomMoral, line 69: Why did you removed these tests?

alphacsc/tests/test_z_encoder.py
```diff
@@ -3,22 +3,13 @@
...
+from conftest import N_ATOMS, N_TIMES_ATOM
 
-@pytest.fixture
-def D_hat(X, rank1):
-    return init_dictionary(X,
-                           N_ATOMS,
-                           N_TIMES_ATOM,
-                           random_state=0,
-                           rank1=rank1)
+N_CHANNELS, N_TIMES = 3, 30
```
tomMoral, line 12: Why not import the one from conftest?

tomMoral (commented):
I did a bunch of comments but overall I am happy with the proposed changes! thx a lot @hndgzkn for all the hard work.

alphacsc/_d_solver.py
hndgzkn: Changed.

alphacsc/_z_encoder.py
hndgzkn: Moved update_z_hat contents to set_D.

alphacsc/tests/test_learn_d_z_multi.py
hndgzkn: I modified it to exclude `auto` values. That way we can use it in more places and avoid unnecessary testing for `auto` values. 

On the other hand, I wanted to make sure that those `auto` values are correctly converted in `learn_d_z_multi`.

alphacsc/tests/test_z_encoder.py
hndgzkn: As they were different, I did not touch the values. I could have asked actually. 
I import from conftest now.

alphacsc/tests/test_update_d_multi.py
hndgzkn: I moved the function to `DSolver` and tests to `test_d_solver.py`

alphacsc/_d_solver.py
hndgzkn: It checks if `D_init` was an ndarray actually. The same check in `init_dictionary` as `if window and not isinstance(D_init, np.ndarray):`

I've put a comment, let me know if it is ok.

alphacsc/tests/test_learn_d_z_multi.py
tomMoral: Ok, thx for the explanation.

alphacsc/tests/test_update_d_multi.py
tomMoral: ok, sorry for the noise, I did not notice.

alphacsc/_d_solver.py
```diff
@@ -124,44 +222,16 @@ def init_dictionary(self, X, n_atoms, n_times_atom, D_init=None,
...
-
-        D_hat, pobj = fista(
-            self._get_objective(z_encoder), self._get_grad(z_encoder),
-            self._get_prox(z_encoder), None, D_hat0, self.max_iter,
-            momentum=self.momentum, eps=self.eps, adaptive_step_size=True,
-            name=self.name, debug=self.debug, verbose=self.verbose
-        )
+        # if D_init is not an ndarray
+        # when D_init is an ndarray, the strategy has D_init attribute
+        if not hasattr(self.init_strategy, 'D_init'):
```
tomMoral, line 229: Maybe better to check this with the type of `IdentityStrategy`?
```suggestion
        # Only use window on D_hat if D_init is not an ndarray
        if not isinstance(self.init_strategy, IdentityStrategy):
```

alphacsc/_z_encoder.py
```diff
@@ -372,20 +358,21 @@ def get_max_error_patch(self):
...
-    def set_reg(self, reg):
-        self.reg = reg
+        nb_missing_atoms = D.shape[0] - self.z_hat.shape[1]
 
-    def add_one_atom(self, new_atom):
-        assert new_atom.shape == (self.n_times_atom + self.X.shape[1],)
-        self.D_hat = np.concatenate([self.D_hat, new_atom[None]])
-        self.z_hat = np.concatenate(
-            [self.z_hat, self._get_new_z_hat(1)], axis=1
-        )
```
tomMoral, line 388: Maybe add a check that `nb_missing >= 0`? This should never be the opposite.

alphacsc/tests/test_learn_d_z_multi.py
```diff
@@ -95,74 +101,64 @@ def test_learn_d_z_multi_dicodile(window):
...
                   uv_constraint=uv_constraint, solver_d=solver_d, rank1=rank1,
                   random_state=0, n_iter=1, solver_z='l-bfgs', D_init=D_init)
+
     res_True = learn_d_z_multi(window=True, **kwargs)
 
     assert not np.allclose(res_False[2], res_True[2])
 
 
-def test_online_learning():
+@pytest.mark.parametrize('n_trials', [N_TRIALS])
```
tomMoral, line 130: Maybe better to put `N_TRIALS` as a constant by default and only parametrize the test if multiple values are passed?

alphacsc/tests/conftest.py
```diff
@@ -38,6 +35,18 @@ def X(rng, n_trials):
     return rng.randn(n_trials, N_CHANNELS, N_TIMES)
```
tomMoral, line 35: Is it possible to have `n_trials=N_TRIALS` in the fixture, in order to avoid the extra parametrization?

tomMoral (approved):
A few more comments but overall I think we can merge once they are adressed.

alphacsc/_d_solver.py
hndgzkn: I think it would be better to avoid passing concrete strategy class name to `d_solver`. It does not have to know about class types. 

As a solution I've set `D_init` as class variable in `d_solver`. And I check as in original code, if it is of type `ndarray`. Another option would be to set it in `strategy` class.

alphacsc/_z_encoder.py
hndgzkn: Done!

alphacsc/init_dict.py
```diff
@@ -18,45 +18,254 @@
...
+        A seed to generate a RandomState instance or the instance itself.
+    """
+
+    def __init__(self, shape, random_state):
+
+        self.shape = shape
+        self.rng = check_random_state(random_state)
+
+    def initialize(self, X):
+        return self.rng.randn(*self.shape)
```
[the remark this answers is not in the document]
hndgzkn: Ok. I've updated. I thought it would be good to have different dictionaries each time the function is called.

alphacsc/tests/conftest.py
hndgzkn: Unfortunately it does not work. I had tried it before.

But instead I created a fixture `def n_trials()` and it worked.

alphacsc/tests/test_learn_d_z_multi.py
hndgzkn: Fixed, see next comment.

alphacsc/tests/conftest.py
tomMoral: Nice one!

alphacsc/init_dict.py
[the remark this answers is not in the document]
tomMoral: This can be achieve by passing a `rng` as `random_state` in the init.
But in general, I find it less surprising like this, `random_state` fixes the output for `initialize`.

tomMoral (approved):
LGTM! thx a lot @hndgzkn, this makes the code much cleaner.

alphacsc/learn_d_z_multi.py
```diff
@@ -295,20 +283,18 @@ def _batch_learn(z_encoder, d_solver, end_iter_func, n_iter=100,
...
 
-        if lmbd_max not in ['fixed', 'scaled']:
-            reg_ = reg * get_lambda_max(X, z_encoder.D_hat)
-            if lmbd_max == 'shared':
-                reg_ = reg_.max()
-            z_encoder.set_reg(reg_)
+        if lmbd_max in ['per_atom', 'shared'] or (
+                lmbd_max == 'scaled' and ii == 0
+        ):
+            z_encoder.update_reg()
```
tomMoral, line 293: Here, `per_atom` is not respected anymore no?
I propose to either remove it or to pass `lmbda_max` as an argument?

hndgzkn: I did not undestand. 

There was `if lmbd_max=='shared'` line, do you mean that?

tomMoral: yes

tomMoral: That's fine like this, thx!

tomMoral:
Thx @hndgzkn for the nice refactoring!

PR #63, opened 2022-01-19
MNT remove wheels for 32bit configs

deepcharles:
`cibuilwheel` freezes when testing wheels for certain 32-bits configurations. Not sure why but it also happened [here](https://github.com/scikit-learn/scikit-learn/issues/22121#issue-1093198865) and [there](https://github.com/deepcharles/ruptures/pull/225#issuecomment-1015301852).

This PR skips the wheel building process for manylinux_i686 (py38 and 39) which are the problematic configurations.


tomMoral (approved):
Let's see if this work for the release!

PR #64, opened 2022-01-19
MTN debug wheel builder

tomMoral:


.github/workflows/upload-to-pypi.yml
```diff
@@ -13,7 +13,7 @@ env:
   SETUPTOOLS_SCM_PRETEND_VERSION: ${{ github.event.inputs.overrideVersion }}
   # Run the package tests using `pytest`
   CIBW_TEST_REQUIRES: pytest
-  CIBW_TEST_COMMAND: pytest --pyargs alphacsc
+  CIBW_TEST_COMMAND: pytest --pyargs alphacsc -k "not sporco and not kmc2"
```
deepcharles, line 16: It seems that wheels are tested with tests that are outside the `alphacsc/tests/` folder (for instance tests from `alphacsc/other/`).

Maybe try
```suggestion
  CIBW_TEST_COMMAND: pytest --pyargs alphacsc/tests
```

deepcharles: wdyt @tomMoral

tomMoral: Yes removing sporco is a good idea. It is probably ok to remove it.
Would it be okay to do the two changes in this PR?

hndgzkn:
The job `build_wheels` fails on all `musllinux` images.  As far as I understand, `scipy` doesn't have images for `musllinux` so tries to rebuild them from source and fails. 

All `many_linux_i686` and `many_linux_x86_64` images all work on my local. 

For `many_linux_aarch64`, the docker image immediately stops on my local, I could not find the reason, so I cannot debug locally. 
On CI machine as far as I understand, the wheels for `numba` `llvmlite` are missing for Python 3.6 and it tries to rebuild and fail. 

I would propose to remove all `musllinux` and try versions above `3.7` for `aarch64`. What do you think @deepcharles @tomMoral ?

deepcharles:
> I would propose to remove all musllinux and try versions above 3.7 for aarch64.

`musllinux` is also a problem for another project I work on, so I'm ok with that

tomMoral:
Seems to work now!!
image

Super cool! thanks @hndgzkn

tomMoral (commented):
LGTM! :tada:

PR #67, opened 2022-03-01
ENH Create docs for multiple release versions

hndgzkn:
Uses `sphinx-multiversion` to create the docs for each release.


agramfort:
@hndgzkn you need my help to debug? if not can you share a rendered version?

hndgzkn:
> @hndgzkn you need my help to debug? if not can you share a rendered version?

I think it is due to brach/tag settings in `conf.py` file. The error is:

`No matching refs found!`

What I have in local is:

multi

agramfort:
we should have a build of the doc done by the CIs right? it's broken here?

hndgzkn:
> we should have a build of the doc done by the CIs right? it's broken here?

Yes.

agramfort:
you know how to fix?

hndgzkn:
> you know how to fix?

I've set `fetch-depth: 0` in `checkout` github action to fetch all history for all branches and tags. See https://github.com/actions/checkout . It can get the refs now. CI started building the docs. It will take some time, then we can check the result.

hndgzkn:
> it's failing now cause it's taking too much time?

No, it is due to running sphinx-build with MPI. I try to repeat it locally, but I have a problem installing MPI.

agramfort:
I am no MPI wizard sadly :(

hndgzkn:
> I am no MPI wizard sadly :(

`sphinx-multiversion` starts processes using `subprocess`. It causes problem with MPI. 

I found [this issue](https://github.com/open-mpi/ompi/issues/3158), I am trying to understand how I can fix the problem.

PR #75, opened 2022-03-21
Update simulate.py

dinalzein:
The function load_data in simulate.py with f_noise=False returns an error: 'tuple' object cannot be interpreted as an integer. I fixed it.


agramfort (approved):
LGTM

for some reason there are no tests in alphacsc/datatests/tests/test_simulate.py

@hndgzkn @rprimet may I put this on your todo list?

thx @dinalzein

agramfort:
thx @dinalzein

PR #80, opened 2022-03-23
CLN remove dead code

tomMoral:


agramfort:
CIs are resurrected ! let's see how red this is ...

tomMoral:
Why would this be red? ^^ I checked the tests and it was working locally :)

agramfort (approved):
green !

@hndgzkn @rprimet I let you have a look to follow and merge if happy

thx @tomMoral

hndgzkn:
It was quite a cleaning :)

We can also remove `pooch` and `tqdm` [here](https://github.com/tomMoral/alphacsc/blob/855d55bc464e726a7dd35742ed33f7fdc684f8f9/setup.cfg#L48).

And remove [this line](https://github.com/tomMoral/alphacsc/blob/855d55bc464e726a7dd35742ed33f7fdc684f8f9/.github/workflows/deploy_ghpages.yml#L25).

We have put them, when mne did not install these dependencies. With latest release of mne, they are no more necessary.

If I have rights to push and it's ok for you, I can make modifications.

rprimet:
Nice!

.github/workflows/deploy_ghpages.yml
```diff
@@ -22,7 +22,6 @@ jobs:
             apt-get install -y gcc git libopenmpi-dev openmpi-bin
             python -m pip install --upgrade pip
             python -m pip install -e .[doc,dicodile]
-            python -m pip install git+https://github.com/mne-tools/mne-python.git@main
```
tomMoral, line 25: But with this, we don't install mne anymore no?

hndgzkn: mne exists in `install_requires` section in `setup.cfg`.

tomMoral: you are right, sorry for the noise.

PR #83, opened 2022-03-29
Update bench_methods_run.py

dinalzein:
learn_d_z function returns 5 variables instead of 4
uv_constraint should be 'auto' because rank1=False


tomMoral (approved):
LGTM, thx @dinalzein .

These files were not maintained with the recent API changes, if you see any other inconsistencies, feel free to correct them and send a PR.

agramfort (approved):

agramfort:
Thx @dinalzein

PR #87, opened 2022-05-04
Multiversion doc

hndgzkn:
Changes sphinx theme from sphinx-bootstap to sphinx-pydata-theme to display documentation for every release.

This requires some changes in
*  https://github.com/alphacsc/alphacsc.github.io repo:
    - a version.txt file in root directory of the project that contains the latest stable version
    - a dev directory that contains documentation for development version
    - a stable directory that contains documentation for the latest release
    - versionwarning.js file that displays a warning and link to stable version on versions other than stable 

* release process:
   - before each release doc/_static/switcher.json file should be updated with new release version.
   
Documentation generation process:
- For each push/pr merge to master branch or release candidate tag with 'rc' generated documentation in pushed to dev directory under documentation repo.
- For each release generated documentation is pushed to stable directory.

### DIFF
--- a/.github/workflows/deploy_ghpages.yml
+++ b/.github/workflows/deploy_ghpages.yml
@@ -42,7 +42,7 @@ jobs:
           path: doc/_build/html/
       - name: Commit to documentation repo
         run: |
-          git clone --no-checkout https://github.com/hndgzkn/hndgzkn.github.io.git --branch master --single-branch gh-pages
+          git clone --no-checkout https://github.com/hndgzkn/hndgzkn.github.io.git --branch main --single-branch gh-pages
           [ ! -d "gh-pages/dev" ] && mkdir -p gh-pages/dev
           rm -r gh-pages/dev/*
           cp -r doc/_build/html/* gh-pages/dev
@@ -56,6 +56,6 @@ jobs:
         uses: ad-m/github-push-action@v0.6.0
         with:
           repository: hndgzkn/hndgzkn.github.io
-          branch: master
+          branch: main
           directory: gh-pages
           github_token: ${{ secrets.DEPLOY_PAGES }}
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy_ghpages.yml
```diff
@@ -1,16 +1,32 @@
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - name: Cache documentation
+        if: ${{ github.ref_type }} != 'tag'
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-documentation
+        with:
+          path: doc/_build/
```
tomMoral, line 26: I think you mainly need to cache the example no?
```suggestion
          path: |
            doc/_build/
            doc/auto_examples
```

tomMoral (approved):
LGTM. If my suggestion fixes the cache, lets use it, else, let's move on and fix the cache latter on.

.github/workflows/deploy_ghpages.yml
hndgzkn: Yes, actually I was thinking that sphinx used _build/doctree directory, probably your change will work then. Thanks!

Shall I update the [documentation repo](https://github.com/alphacsc/alphacsc.github.io) then?

tomMoral: If this is working, feel free to change the doc and merge :)

hndgzkn: Unfortunately cache still does not work :(

tomMoral:
The cache is now working! 🎉

agramfort (approved):
Nice ! Merge if happy @hndgzkn

hndgzkn:
> The cache is now working! tada

Yes, it is working :) 

But there is a problem with deploy job. I try to understand the reason. There are recent changes in [Push action](https://github.com/ad-m/github-push-action) that we use. I try to understand if it is due to these changes.

hndgzkn:
Probably not as we pinned to version 0.6.0 :(

hndgzkn:
Ok. It was due to incorrect if condition in deploy job. I fixed it. I am merging

PR #89, opened 2022-05-10
Fix building wheels

hndgzkn:
Updates build jobs not to use cibuildwheels as all cython code is removed.

### DIFF
--- a/.github/workflows/deploy_ghpages.yml
+++ b/.github/workflows/deploy_ghpages.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Cache documentation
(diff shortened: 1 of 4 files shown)


tomMoral (approved):
This looks much simpler. You have tested that this gives an installable package with a rc?

hndgzkn:
> This looks much simpler. You have tested that this gives an installable package with a rc?

No, I haven't yet. I wanted to get your comments before.

In that case, I try with a rc?

tomMoral:
I have to say I am not super familiar with this so as long as this works well, I am fine with it.
We just need to make sure we correctly remove unnecessary steps. Feel free to merge once you think it is working

PR #99, opened 2022-09-16
Fix benchmark bug

hndgzkn:
- Fixes segment counts
- Simplifies benchmark methods
- Modifies benchmark plot to group plots depending on channel_count and rank1 status


alphacsc/utils/coordinate_descent.py
```diff
@@ -56,15 +56,15 @@ def _coordinate_descent_idx(Xi, D, constants, reg, z0=None, max_iter=1000,
...
+            n_seg = max(n_times_valid // (2 * n_times_atom - 1), 1)
         elif strategy in ('random', 'cyclic'):
             n_seg = 1
             n_coordinates = n_times_valid * n_atoms
 
     rng = check_random_state(random_state)
 
     max_iter *= n_seg
-    n_times_seg = n_times_valid // n_seg + 1
+    n_times_seg = n_times_valid // n_seg
```
tomMoral, line 68: i think the test are failing because with this values, you don't cover all the coordinate with the segments.
You need to have enough segment so that `n_seg * n_times_seg > n_times_valid`.

alphacsc/utils/coordinate_descent.py
```diff
@@ -95,7 +95,9 @@ def objective(zi):
     i_seg = 0
     seg_bounds = [0, n_times_seg]
     t0, k0 = -1, 0
+
     for ii in range(int(max_iter)):
+        i_seg = i_seg % n_seg
```
tomMoral, line 100: You should not need this as this is handled by line 149-152

benchmarks/bench_methods_run.py
```diff
@@ -84,90 +86,52 @@ def run_l_bfgs(X, ds_init, reg, n_iter, random_state, label, factr_d=1e7,
...
 
-def run_multichannel_gcd(X, ds_init, reg, n_iter, random_state, label):
-    if X.ndim == 2:
-        n_atoms, n_times_atom = ds_init.shape
-        ds_init = np.c_[np.ones((n_atoms, 1)), ds_init]
-        X = X[:, None, :]
-    else:
-        n_atoms, n_channels, n_times_atom = ds_init.shape
-        ds_init = get_uv(ds_init)  # project init to rank 1
+def run_multivariete(X, solver_z, reg, n_iter, random_state, label, rank1,
```
tomMoral, line 89:
```suggestion
def run_multivariate(X, solver_z, reg, n_iter, random_state, label, rank1,
```

benchmarks/bench_methods_run.py
```diff
@@ -84,90 +86,52 @@ def run_l_bfgs(X, ds_init, reg, n_iter, random_state, label, factr_d=1e7,
...
-    pobj, times, d_hat, z_hat, reg = learn_d_z_multi(
-        X, n_atoms, n_times_atom, solver_d='auto', solver_z="dicodile",
-        uv_constraint='auto', eps=-np.inf, solver_z_kwargs=solver_z_kwargs,
-        reg=reg, solver_d_kwargs=dict(max_iter=100), n_iter=n_iter,
-        random_state=random_state, raise_on_increase=False, D_init=ds_init,
-        n_jobs=njobs, verbose=verbose, rank1=True)
 
-    # remove the ds init duration
-    times[0] = 0
+    return run_multivariete(X, "lgcd", reg, n_iter, random_state, label,
```
tomMoral, line 110:
```suggestion
    return run_multivariate(X, "lgcd", reg, n_iter, random_state, label,
```

benchmarks/bench_methods_run.py
```diff
@@ -84,90 +86,52 @@ def run_l_bfgs(X, ds_init, reg, n_iter, random_state, label, factr_d=1e7,
...
 
+def run_multichannel_dicodile(X, reg, n_iter, random_state, label, njobs=30):
+    if X.ndim == 2:
+        X = X[:, None, :]
 
-def run_multichannel_gcd_fullrank(X, ds_init, reg, n_iter, random_state,
-                                  label):
-    assert X.ndim == 3
-    n_atoms, n_channels, n_times_atom = ds_init.shape
+    return run_multivariete(X, "dicodile", reg, n_iter, random_state, label,
```
tomMoral, line 118:
```suggestion
    return run_multivariate(X, "dicodile", reg, n_iter, random_state, label,
```

benchmarks/bench_methods_run.py
```diff
@@ -84,90 +86,52 @@ def run_l_bfgs(X, ds_init, reg, n_iter, random_state, label, factr_d=1e7,
...
-        random_state=random_state, raise_on_increase=False, D_init=ds_init,
-        n_jobs=n_jobs, verbose=verbose, rank1=False)
 
-    # remove the ds init duration
-    times[0] = 0
+def run_multichannel_gcd_fullrank(X, reg, n_iter, random_state, label):
+    assert X.ndim == 3
 
-    return pobj[::2], np.cumsum(times)[::2], d_hat, z_hat
+    return run_multivariete(X, "lgcd", reg, n_iter, random_state, label, False,
```
tomMoral, line 125:
```suggestion
    return run_multivariate(X, "lgcd", reg, n_iter, random_state, label, False,
```

benchmarks/bench_methods_run.py
```diff
@@ -84,90 +86,52 @@ def run_l_bfgs(X, ds_init, reg, n_iter, random_state, label, factr_d=1e7,
...
-        uv_constraint='auto', eps=-np.inf, solver_z_kwargs=solver_z_kwargs,
-        reg=reg, solver_d_kwargs=dict(max_iter=100), n_iter=n_iter,
-        random_state=random_state, raise_on_increase=False, D_init=ds_init,
-        n_jobs=njobs, verbose=verbose, rank1=False)
-
-    # remove the ds init duration
-    times[0] = 0
 
-    return pobj[::2], np.cumsum(times)[::2], d_hat, z_hat
+    return run_multivariete(X, "dicodile", reg, n_iter, random_state, label,
```
tomMoral, line 133:
```suggestion
    return run_multivariate(X, "dicodile", reg, n_iter, random_state, label,
```

benchmarks/bench_methods_run.py
```diff
@@ -220,28 +183,17 @@ def one_run(X, X_shape, random_state, method, n_atoms, n_times_atom, reg):
...
     else:
         n_trials, n_channels, n_times = X.shape
-        X_init = X
-
-    # use the same init for all methods
-    ds_init = init_dictionary(X_init, n_atoms, n_times_atom, D_init='chunk',
-                              rank1=False, uv_constraint='separate',
-                              random_state=random_state)
-    if len(X_shape) == 2:
-        ds_init = ds_init[:, 0, :]
```
tomMoral, line 234: Quick question is: are you sure the init is the same for all the methods (same random_state)?
If so, works for me.

tomMoral (commented):
A few changes required to make sure you cover all segments but other wise it seems nice.

benchmarks/bench_methods_run.py
hndgzkn: Yes, at line 250 random_state is set `iterator = itertools.product(methods, range(n_states))`. 
All methods use the same random_state and I checked the dict to see if the values are the same.

alphacsc/utils/coordinate_descent.py
hndgzkn: Ok. I understand. I had tried to make segment counts equal with dicodile. I did not pay attention to that. I've reverted and fixed on dicodile side (If I understand correct). 

I want to add a test to see that dicodile and alphacsc gives the same values after each iteration.

alphacsc/_d_solver.py
```diff
@@ -428,13 +427,21 @@ def update_D(self, z_encoder):
         # use FISTA on alternate u and v
 
         pobj = []
+        X = z_encoder.X
+        z_hat = z_encoder.get_z_hat()
```
tomMoral, line 431: You should not need `z_hat` if you have `loss=='l2'`.

alphacsc/_d_solver.py
```diff
@@ -324,7 +324,9 @@ def update_D(self, z_encoder):
         D_hat0 = self._windower.remove_window(self.D_hat)
 
         D_hat, pobj = fista(
-            self._get_objective(z_encoder), self._get_grad(z_encoder),
+            self._get_objective(z_encoder),
+            self._get_grad(z_encoder.X, z_encoder.get_z_hat(),
+                           z_encoder.get_constants()),
```
tomMoral, line 329:
```suggestion
            self._get_objective(z_encoder), self._get_grad(z_encoder),
```

alphacsc/_d_solver.py
```diff
@@ -194,13 +194,13 @@ def prox(D, step_size=None):
 
         return prox
 
-    def _get_grad(self, z_encoder):
+    def _get_grad(self, X, z_hat, constants):
```
tomMoral, line 197: something like this?
```suggestion
    def _get_grad(self, X, z_encoder):
        constants = z_encoder.get_constants()
        z_hat = z_encoder.get_z_hat() if self.loss == "l2" else None
```

tomMoral: At least we should only call `get_z_hat` here, not in `update_D` and follow up `update_u/v`.

tomMoral (commented):
I would say, I agree that we should avoid calling `get_z_hat` multiple times but we should still pass `z_encoder` as much as we can, to avoid code that is too verbose on arguments.
I feel that right now, this is not the right level.

alphacsc/utils/coordinate_descent.py
```diff
@@ -54,17 +54,18 @@ def _coordinate_descent_idx(Xi, D, constants, reg, z0=None, max_iter=1000,
...
 
+    n_times_seg = 2 * np.array(n_times_atom) - 1
     if n_seg == 'auto':
         if strategy == 'greedy':
-            n_seg = max(n_times_valid // (2 * n_times_atom), 1)
+            n_seg = max(1, n_times_valid // n_times_seg +
+                        ((n_times_valid % n_times_seg) != 0))
         elif strategy in ('random', 'cyclic'):
             n_seg = 1
             n_coordinates = n_times_valid * n_atoms
```
tomMoral, line 65: This code will break if `n_seg` is not `'auto'`, as `n_times_seg` won't be correct.
Could you add a `else` that compute `n_times_seg` when `n_seg` is given by the user:
```suggestion
    else:
        n_times_seg = n_times_valid // n_seg + (n_times_valid % n_seg != 0)
```

benchmarks/bench_methods_plot.py
```diff
@@ -278,11 +288,13 @@ def plot_barplot(all_results_df, threshold, normalize_method, save_name):
...
-        plot_convergence(data_frame, threshold=1e-2, normalize_method='last',
+        plot_convergence(data_frame, threshold=threshold,
+                         normalize_method=normalize_method,
                          save_name=load_name[:-4])
         plt.close('all')
 
     # plot the aggregation of all results
-    plot_barplot(all_results_df, threshold=1e-2, normalize_method='last',
+    plot_barplot(all_results_df, threshold=threshold,
+                 normalize_method='last',
```
tomMoral, line 298: I am not a big fan of `normalize_method='last'`. Could you use `best` or `none`?

hndgzkn: I've changed it, however I realised that for `None`, barplot returns None. 

And for `best`, it raises exception:
```
Traceback (most recent call last):
  File "bench_methods_plot.py", line 299, in <module>
    save_name=str(figures_dir / 'all'))
  File "bench_methods_plot.py", line 189, in plot_barplot
    pobj = normalize_pobj(pobj, None, normalize_method)
  File "bench_methods_plot.py", line 31, in normalize_pobj
    assert best_pobj is not None
AssertionError

```

hndgzkn:
> Could you re-run the benchmark? I suspect the fact that we don't communicate `z_hat` anymore will make it much faster for dicodile1?

It makes some difference for 1 worker, it is ~4 times slower instead of ~5 times. For more workers, it makes no difference.
all_last_001_T_269400_P__5_K__2_L__32_rank1__False

hndgzkn:
Yes, I've tried with eps and tol that I have not used before, it gives similar results.

tomMoral (approved):

PR #101, opened 2022-10-14
Fix doc generation

hndgzkn:
After new release of pydata-sphinx-theme doc building does not work anymore.

- Fixes json URL path
- Fixes docstrings that give warning during doc building

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -115,7 +115,7 @@ html_theme_options = {
     "show_toc_level": 1,
     "navbar_end": ["version-switcher", "navbar-icon-links"],
     "switcher": {
-        "json_url": "/stable/_static/switcher.json",
+        "json_url": "https://alphacsc.github.io/stable/_static/switcher.json",
         "version_match": version_match,
     },
 }
--- a/alphacsc/utils/dictionary.py
+++ b/alphacsc/utils/dictionary.py
@@ -11,8 +11,8 @@ def get_D(uv_hat, n_channels):
     n_channels: int
         number of channels in the original multivariate series
 
-    Return
-    ------
+    Returns
+    -------
     D: array (n_atoms, n_channels, n_times_atom)
     """
 
@@ -26,8 +26,8 @@ def get_uv(D):
     ---------
     D: array (n_atoms, n_channels, n_times_atom)
 
-    Return
-    ------
+    Returns
+    -------
     uv: array (n_atoms, n_channels + n_times_atom)
     """
     n_atoms, n_channels, n_times_atom = D.shape
--- a/alphacsc/utils/signal.py
+++ b/alphacsc/utils/signal.py
@@ -40,8 +40,8 @@ def split_signal(X, n_splits=1, apply_window=True):
         reduce the border artifacts by reducing the weights of the chunk
         borders.
 
-    Return
-    ------
+    Returns
+    -------
     X_split: ndarray, shape (n_splits, n_channels, n_times // n_splits)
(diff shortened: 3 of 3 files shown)


doc/conf.py
```diff
@@ -115,7 +115,7 @@
     "show_toc_level": 1,
     "navbar_end": ["version-switcher", "navbar-icon-links"],
     "switcher": {
-        "json_url": "/stable/_static/switcher.json",
+        "json_url": "https://alphacsc.github.io/stable/_static/switcher.json",
```
tomMoral, line 118: Maybe better to take the one from master?
```suggestion
        "json_url": "https://raw.githubusercontent.com/alphacsc/alphacsc/master/doc/_static/switcher.json",
```
or
```suggestion
        "json_url": "https://alphacsc.github.io/dev/_static/switcher.json",
```

hndgzkn: The default page that we display for docs is stable. So it seems more natural for me to take version info from stable. 

Besides for each push to master, dev is rebuilt and deployed. It is not likely, but if a problem occurs in dev, it will not be affected from that.  

I'd prefer to keep as stable. What do you think?

tomMoral: The thing that I don't understand is that you need to make a release in order to change this?
For instance what happens if you want to add a rc version in the switcher?

Sorry I am probably overfitting. If you think it is not a problem as this should only be changed when we release a nez version, fine for me.

hndgzkn: For release candidates, we do not create a documentation version. It is deployed under dev. We had decided to it that way, if you prefer otherwise, I can change it. But currently, CI job does not let you create a version for RCs.

tomMoral: What I am saying is that if we do a release (`0.5` for instance), the documentation will not be updated automatically.
`stable` will still point to `0.4` as the `switcher.json` won't be up to date (it will be the one currently exposed).
So it seems that this pattern do not allow for updates? but maybe I am wrong?

If you take the `dev` one however, it will be the current one on `master`, so we can update it before doing a release.

hndgzkn: Below is from [CI job](https://github.com/hndgzkn/alphacsc/blob/c9ecf7fd13ec8adc711e9f10fdd304535aa5f2e3/.github/workflows/deploy_ghpages.yml#L48) where docs are deployed:

```
         # copy documentation to dev directory
          # when it is not a tag or a it is a tag but release candidate
          if [ ${{github.ref_type}} != 'tag' ] || [[ ${{github.ref_name}} =~ "rc" ]]; then
              rm -Rf gh-pages/dev
              cp -a doc/_build/html/ gh-pages/dev
          fi;
          # if release:
          # rename current stable directory with the version in version.txt
          # copy documentation under stable directory
          # update version.txt with current version
          if [ ${{github.ref_type}} == 'tag' ] && [[ ! ${{github.ref_name}} =~ "rc" ]]; then
             mv gh-pages/stable gh-pages/$(cat gh-pages/version.txt)
             cp -a doc/_build/html/ gh-pages/stable
             echo ${{github.ref_name}} > gh-pages/version.txt
          fi;
```
For example when we release `0.5`, we need to update `switcher.json` before release to have version `0.5` inside. When we make the release, existing contents of `stable` will be moved to `0.4` directory and latest built documentation will be put under `stable` so will have latest version of `switcher.json`.

tomMoral: Ok, I understand now. You are right then. Thanks taking the time to explain it to me.

PR #102, opened 2022-10-26
add data loader for camcan dataset + add event description into info

CedricAllain:
1. `alphacsc/datasets/camcan.py`: add a data loader function with default pre-processing parameters
2. `alphacsc/dataset/mne_data.py`: add an event description dictionary in `info['event_info']`

### DIFF
--- a/alphacsc/datasets/camcan.py
+++ b/alphacsc/datasets/camcan.py
@@ -1,0 +1,180 @@
+""" Cambridge Centre for Ageing and Neuroscience (CamCAN) dataset
+Data available at: https://www.cam-can.org/index.php?content=dataset
+
+Shafto, M.A., Tyler, L.K., Dixon, M., Taylor, J.R., Rowe, J.B., Cusack, R.,
+Calder, A.J., Marslen-Wilson, W.D., Duncan, J., Dalgleish, T., Henson, R.N.,
+Brayne, C., CamCAN, & Matthews, F.E. (2014). The Cambridge Centre for Ageing
+and Neuroscience (CamCAN) study protocol: a cross-sectional, lifespan,
+multidisciplinary examination of healthy cognitive ageing. BMC Neurology,
+14(204). doi:10.1186/s12883-014-0204-1
+https://bmcneurol.biomedcentral.com/articles/10.1186/s12883-014-0204-1 
+"""
+import os
+from os.path import join
+from copy import deepcopy
+
+import mne
+import numpy as np
+import pandas as pd
+from joblib import Memory
+
+from mne_bids import BIDSPath, read_raw_bids
+
+from alphacsc.utils.signal import split_signal
+
+DATA_DIR = "/storage/store/data/"
+PARTICIPANTS_FILE = join(DATA_DIR, "camcan/BIDSsep/smt/participants.tsv")
+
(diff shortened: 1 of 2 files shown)


alphacsc/datasets/camcan.py
```diff
@@ -0,0 +1,202 @@
...
+        .
+        ├── camcan
+        │   └── BIDSsep
+        │       └── smt
+        │           ├── participants.tsv  # subjects' informations
+        │           ├── sub-CC110033      # one subject's folder
+        │           └── ...
+        └── camcan-mne
+            ├── Cam-CAN_sss_cal.dat       # calibration file
+            └── Cam-CAN_ct_sparse.fif     # cross_talk file
```
agramfort, line 79: this is specific to our setup at Inria

alphacsc/datasets/camcan.py
```diff
@@ -0,0 +1,202 @@
...
+
+from mne_bids import BIDSPath, read_raw_bids
+
+try:
+    from ..utils.signal import split_signal
+except ValueError:
+    from alphacsc.utils.signal import split_signal
+
+
+DATA_DIR = "/storage/store/data/"
```
agramfort, line 29: alphacsc is meant to be used by anyone not just us at Inria

alphacsc/datasets/camcan.py
```diff
@@ -0,0 +1,202 @@
...
+PARTICIPANTS_FILE = join(DATA_DIR, "camcan/BIDSsep/smt/participants.tsv")
```
agramfort, line 30: eventually the BIDSsep folder could be referred to as bids_root

alphacsc/datasets/mne_data.py
```diff
@@ -68,6 +69,9 @@ def load_data(dataset="somato", n_splits=10, sfreq=None, epoch=None,
         raw = mne.io.read_raw_fif(file_name, preload=True)
         raw.notch_filter(np.arange(60, 181, 60), n_jobs=n_jobs)
         event_id = [1, 2, 3, 4]
+        event_des = {'auditory/left': 1, 'auditory/right': 2,
+                     'visual/left': 3, 'visual/right': 4,
+                     'smiley': 5, 'buttonpress': 32}
```
agramfort, line 74: this is an event_id for mne

agramfort (commented):
We need to discuss where to put this code as alphacsc is probably not a great fit

alphacsc/datasets/camcan.py
CedricAllain: I tried to mimic and reproduced how paths were handled in `alphacsc/datasets/hcp.py`

alphacsc/datasets/mne_data.py
CedricAllain: I updated `event_id` to be this dict.

CedricAllain:
> We need to discuss where to put this code as alphacsc is probably not a great fit

Indeed maybe it is not really appropriate, I did this as there is a data loader of HCP in `datasets/hcp.py`.

agramfort:
to me the hcp loader should not be present here or at least not using hard coded path. There is no chance this hcp code is working for anyone.

PR #104, opened 2022-11-14
DOC fix parameters multicsc

tomMoral:


agramfort:
ok much better

🙏 @tomMoral

tomMoral:
It is still 1h vs 20min before, we need to investigate why

PR #105, opened 2022-11-16
ENH support for greedy with dicodile

tomMoral:


agramfort:
any docstring to update or whats new entry to add?

tomMoral:
We agreed with @rprimet that we should also add in this PR an example on `sample` with `dicodile` that is faster on our cluster.

Also, the error in the test seems legit, we should make sure we investigate it @rprimet

rprimet:
Yes the test error seems legit, we might also add dicodile to the test matrix (it was just found at the coverage step which feels weird...)

rprimet:
looks like releasing + bumping the dicodile version fixes that

tomMoral:
Ok, so now, we need to create an example that shows we go faster :)

PR #107, opened 2022-12-06
ENH add test set

CedricAllain:
Add the possibility to have a test set and compute the cost on it at each update of the dictionnary.


agramfort:
@CedricAllain can you merge master branch here so the diff is more readable? thanks

PR #108, opened 2022-12-07
FIX gradient for FISTA solver+add positive argument to update_z_multi

tomMoral:
- The gradient for the FISTA solver was wrong, it relied on the subgradient used in `l-BFGS`. We could either change the prox to a projection or change the gradient to only include the `l2` loss. I choose the later for consistency with `positive=False`.
- Added a `positive=True` argument, so we can also solve the negative case. This does not make the code more complex but it will accomodate different cases so I think it is worth adding.

### DIFF
--- a/.github/workflows/unittests.yml
+++ b/.github/workflows/unittests.yml
@@ -12,8 +12,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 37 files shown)


agramfort (commented):
does it make the benchmark results more reasonable?

PR #114, opened 2023-04-21
ENH ensure copy when signal spliting

CedricAllain:


alphacsc/utils/signal.py
```diff
@@ -58,7 +58,7 @@ def split_signal(X, n_splits=1, apply_window=True):
 
     n_channels, n_times = X.shape
     n_times = n_times // n_splits
-    X_split = X[:, :n_splits * n_times]
+    X_split = X[:, :n_splits * n_times].copy()
```
tomMoral, line 61: Why do you need the copy?

CedricAllain: When we did some experiments, we found that whithout the copy, `X_split` is still linked to `X` so when we normalized it (`X_split \= X_split.std()`, we also modify `X`.

PR #115, opened 2023-04-21
ENH change max function to quantile with equivalent default value q=1

CedricAllain:
In the utils.dictionary.get_lambda_max() function, update so that it is possible to return a given quantile q instead of the max. Default value is put to q=1 so that it is equivalent to the max.

### DIFF
--- a/alphacsc/utils/dictionary.py
+++ b/alphacsc/utils/dictionary.py
@@ -92,7 +92,7 @@ def _patch_reconstruction_error(X, z, D):
                    for diff_i in diff], axis=1)
 
 
-def get_lambda_max(X, D_hat, sample_weights=None):
+def get_lambda_max(X, D_hat, sample_weights=None, q=1):
 
     # univariate case, add a dimension (n_channels = 1)
     if X.ndim == 2:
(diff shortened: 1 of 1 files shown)


alphacsc/utils/dictionary.py
```diff
@@ -92,7 +92,7 @@ def _patch_reconstruction_error(X, z, D):
                    for diff_i in diff], axis=1)
 
 
-def get_lambda_max(X, D_hat, sample_weights=None):
+def get_lambda_max(X, D_hat, sample_weights=None, q=1):
```
agramfort, line 95: @CedricAllain can you add docstring to this function?

CedricAllain: I have updated, hope it is fine.

alphacsc/utils/dictionary.py
```diff
@@ -92,8 +92,33 @@ def _patch_reconstruction_error(X, z, D):
...
+    """For each atom, compute the associated lambda_max as the q-quantile of
+    the energy patches over the signal length.
```
tomMoral, line 97:
```suggestion
    """For each atom, compute the regularization parameter scaling.
    
    This value is usually defined as the smallest value for which 0 is
    a solution of the optimization problem.
    In order to avoid spurious values, this quantity can also be estimated
    as the q-quantile of the correlation between signal patches and the
    atom.
```

alphacsc/utils/dictionary.py
```diff
@@ -92,8 +92,33 @@ def _patch_reconstruction_error(X, z, D):
...
 
+    Parameters
+    ----------
+    X : array, shape (n_trials, n_times) or \
+               shape (n_trials, n_channels, n_times)
```
tomMoral, line 102:
```suggestion
    X : array, shape (n_trials, n_times) or
               shape (n_trials, n_channels, n_times)
```

alphacsc/utils/dictionary.py
```diff
@@ -92,8 +92,33 @@ def _patch_reconstruction_error(X, z, D):
...
+        The data
+
+    D_hat : array, shape (n_atoms, n_channels + n_times_atoms) or \
+                   shape (n_atoms, n_channels, n_times_atom)
+        The atoms
+
+    sample_weights : None | array, shape (n_trials, n_times) or \
+                                   shape (n_trials, n_channels, n_times)
```
tomMoral, line 110:
```suggestion
    D_hat : array, shape (n_atoms, n_channels + n_times_atoms) or
                   shape (n_atoms, n_channels, n_times_atom)
        The atoms

    sample_weights : None | array, shape (n_trials, n_times) or
                                   shape (n_trials, n_channels, n_times)
```

tomMoral (approved):
A  few nitpicks but other wise LGTM.

tomMoral (approved):

tomMoral:
can you fix the linting?

alphacsc/utils/dictionary.py
```diff
@@ -92,8 +92,38 @@ def _patch_reconstruction_error(X, z, D):
                    for diff_i in diff], axis=1)
 
 
-def get_lambda_max(X, D_hat, sample_weights=None):
+def get_lambda_max(X, D_hat, sample_weights=None, q=1):
+    """For each atom, compute the regularization parameter scaling.
+
```
tomMoral, line 97:
```suggestion

```

PR #120, opened 2023-05-20
DOC improve doc building speed?

tomMoral:
Just found out that using `mpiexec -np 1` forced all process to run on a single core.
Trying to untie this to see if this speeds up the doc build :crossed_fingers: 

At least it did on my machine :)


tomMoral:
50min instead of 2h30!!! 🤯

We have now found why it became so slow when we added the dicodile backend!

might interest you @hndgzkn :)

merging as this is a clear improvement

PR #123, opened 2024-10-14
FIX flake8

tomMoral:


tomMoral:
need to update the [codecov upload action](https://github.com/codecov/codecov-action) and setup a token

PR #126, opened 2025-07-28
ENH custom early stopping

tomMoral:
Allow callback to customize early stopping.


alphacsc/learn_d_z_multi.py
```diff
@@ -419,7 +420,12 @@ def get_iteration_func(eps, stopping_pobj, callback, lmbd_max, name, verbose,
                        raise_on_increase):
     def end_iteration(z_encoder, pobj, iteration):
         if callable(callback):
-            callback(z_encoder, pobj)
+            if callback(z_encoder, pobj):
+                if verbose == 1:
+                    print("")
+                print(f"[{name}] Stopping after {iteration + 1} iterations, "
+                      f"after callback returns True.")
```
tomMoral:
```suggestion
                elif verbose > 1:
                    print(f"[{name}] Stopping after {iteration + 1} iterations, "
                          f"after callback returns True.")
```

PR #130, opened 2026-04-14
CSC-l0 implementation

Nanored4498:
This is a WIP pull request about adding CSC-l0 to the library.
The paper could be find here : https://onlinelibrary.wiley.com/doi/full/10.1002/sam.70000

Currently it is implemented as a ZEncoder named `NoOverlapEncoder` and a DSolver named `NoOverlapSolver`.
There are currently designed to be used with `BatchCDL(solver_z="no-overlap", solver_d="no-overlap")`.
But, it faces many issues. The three main issues are:
* CSC-l0 would be better designed by alternatively updating the temporal support $T$ and the dictionary $D$ while `BatchCDL` alternate between updating $z$ and $D$. This causes bad updates of $D$.
* ~~`BatchCDL` starts with a $D$ initialization, while it would be way better to start with a $T$ initialization.~~ *Edit: A new Initialization strategy `NoOverlapStrategy` can be used to initialize the temporal support T and then apply a step of `NoOverlapSolver` to compute D.*
* `BatchCDL` requires to use dense representation of $z$ while it would be better to use a sparse representation

For this reasons, I think it would be better to make an independent implementation of CSC-l0, with new structures and new optimization functions (other than `learn_d_z_multi`).


alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,431 @@
...
+                 uv_constraint, D_init, resample_strategy, window, eps,
+                 max_iter, momentum, random_state, verbose, debug):
+        super().__init__(n_channels, n_atoms, n_times_atom, solver_d,
+                 uv_constraint, D_init, resample_strategy, window, eps,
+                 max_iter, momentum, random_state, verbose, debug)
+        
+        self.D_hat = np.empty((n_atoms, n_channels, n_times_atom))
+        self.Y = None
+
+    def update_D(self, z_encoder, reorder=False):
```
tomMoral: `reorder` should be a parameter of the `DSolver`

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,431 @@
...
+        self.D_mul = np.linalg.norm(self.D_hat, axis=(1, 2))
+        np.divide(1., self.D_mul, out=self.D_mul, where=self.D_mul!=0)
+        self.cost = None
+
+    def get_constants(self):
+        self._compute_dense_z_hat()
+        return super().get_constants()
+
+
+class NoOverlapSolver(BaseDSolver):
```
tomMoral:
```suggestion
class NoOverlapDSolver(BaseDSolver):
```

alphacsc/_z_encoder.py
```diff
@@ -76,183 +78,16 @@ def get_z_encoder_for(X, D_hat, n_atoms, n_times_atom, n_jobs,
             X, D_hat, n_atoms, n_times_atom, n_jobs,
             solver_kwargs, reg
         )
-    else:
-        raise ValueError(f'unrecognized solver type: {solver}.')
-
-
-class BaseZEncoder:
```
tomMoral, line 83: You needed to move this to avoid circular import?

alphacsc/_base.py
tomMoral, line 1: call `_base.py` would be clearer (you have both `encoder` and `decoder`

alphacsc/_no_overlap.py
tomMoral, line 1: Missing comments, think that we will need to maintain the code later :)

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,431 @@
...
+            t = nz_index[trial][i][1]
+            coeff = nz_coeff[trial][i]
+            proj = 0
+            for c in range(C):
+                proj += D[ind, c] @ X[trial, c, t:t+L]
+            E0 += coeff * (coeff * D2[ind] - 2. * proj)
+    return .5 * E0 + penalty * Ereg
+
+@nb.njit(nb.int64(_int32_r(), _int32_r(3), _float64_r(2), _float64_r(3), _float64_r(3)), cache=True, nogil=True)
+def _find_max_error_patch(nnz, nz_index, nz_coeff, D, X):
```
tomMoral: Isn't there a version of this somewhere in the code that could be improved?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,431 @@
...
+            nnz_atom[atom] += 1
+
+MAX_KMEAN_STEPS = 10
+
+@nb.njit(
+    nb.void(
+        _float64_r(3), _int32_r(), _int32_r(3), _float64_w(), _int32_w(), _float64_w(2), _float64_w(3)
+    ), cache=True, nogil=True
+)
+def kmean(X, nnz, nz_index, init_data, updt_data, Y, D):
```
tomMoral: Can't we use `sklearn` here to compute K-means?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,431 @@
...
+            self.fft_fwd = pyfftw.FFTW(
+                self.fft_data[...,:T2],
+                self.fft_out,
+                axes=(-1,), direction='FFTW_FORWARD', flags=('FFTW_ESTIMATE',)
+            )
+            self.fft_bwd = pyfftw.FFTW(
+                self.proj_in,
+                self.proj[...,:T2],
+                axes=(-1,), direction='FFTW_BACKWARD', flags=('FFTW_ESTIMATE',), normalise_idft=False
+            )
```
tomMoral, line 384: As this is only for the init, is it worth it to use `pyfftw` instead of `numpy`?

tomMoral (commented):
Nice PR :)
This is missing a bit of comments on the `_no_overlap` part, and a few design choices are unclear to me.
Also, adding tests would be nice to make sure this runs correctly.

I have for project to have a second format for `z` that would be a `pandas.DataFrame`, to allow for efficient and self-contained sparse representation. Do you think it could be made efficient to work with such representation?
IMO: yes as the `index` represents the  `nnz_index` and the `values` would give the `nz_coeff`. This would simplify representation and representation translation, as `z_sparse = z.query("value != 0")` and `z_plain = z_sparse.reindex(range(40), fill_value=0)` (to adapt to multi-trial/atoms).

alphacsc/_base.py
Nanored4498: renamed in 5eeb9f9

alphacsc/_z_encoder.py
Nanored4498: Yes, exactly

alphacsc/_no_overlap.py
Nanored4498: I have changed the objective function of the ZEncoder such that it is now always possible to use `reorder=True`. So I removed the parameter

alphacsc/_no_overlap.py
Nanored4498: This implementation is designed to work only for non overlapping $z$.

alphacsc/_no_overlap.py
Nanored4498: As discussed, it is not exactly the K-means algorithm we are used to. I will add some comments to described what the function does.

alphacsc/_no_overlap.py
Nanored4498: As we discussed together. It is also used inside `compute_z`.

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
+import numpy as np
+
+from .init_dict import get_init_strategy
+from .loss_and_gradient import compute_objective
+from .utils.dictionary import get_lambda_max, NoWindow
+from .utils.optim import fista
+from .utils.validation import check_random_state
+
+
+class BaseZEncoder:
```
tomMoral, line 10: Can you add a class description?

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+
+    def __init__(self, X, D_hat, n_atoms, n_times_atom, n_jobs,
+                 solver_kwargs, reg):
+
+        self.X = X
+        self.D_hat = D_hat
+        self.n_atoms = n_atoms
+        self.n_times_atom = n_times_atom
```
tomMoral, line 18: Aren't `n_atoms/n_times_atom` redundant with `D_hat` shape?

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+        self.n_trials, self.n_channels, self.n_times = X.shape
+        self.n_times_valid = self.n_times - self.n_times_atom + 1
+
+        self.XtX = np.dot(X.ravel(), X.ravel())
+
+    def compute_z(self):
+        """
+        Perform one incremental z update.
+        This is the "main" function of the algorithm.
+        """
```
tomMoral, line 33:
```suggestion
        """Encode X by computing z for the current dictionary D_hat.
        
        This is the "main" method of this class.
        """
```

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+        raise NotImplementedError()
+
+    def compute_z_partial(self, i0):
+        """
+        Compute z on a slice of the signal X, for online learning.
```
tomMoral, line 38:
```suggestion
        """
        Encode a slice of the signal X, for online learning.
```

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+        -------
+        D_k : ndarray, shape (n_channels, n_times_atom) or
+                (n_channels + n_times_atom,)
+            Patch of the residual with the largest error.
+        """
+        raise NotImplementedError()
+
+    def get_z_hat(self):
+        """
+        Returns the sparse codes of the signals.
```
tomMoral, line 113:
```suggestion
        Returns z_hat, corresponding to the encoded signal X.
```

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+        Returns
+        -------
+        z_nnz : ndarray, shape (n_atoms,)
+            Ratio of non-zero activations for each atom.
+        """
+        raise NotImplementedError()
+
+    def set_D(self, D):
+        """
+        Update the dictionary.
```
tomMoral, line 146:
```suggestion
        """Update the encoder parameters, usually a dictionary.
```

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+
+        """
+        self.reg = self.reg * get_lambda_max(self.X, self.D_hat)
+
+        if not is_per_atom:
+            self.reg = self.reg.max()
+
+    def get_constants(self):
+        """
+        """
```
tomMoral, line 173:
```suggestion
        """Return information necessary to update the encoder parameter.
        
        For dictionary learning, this corresponds to the signal gram-matrix
        $X^TX$, and the sufficient statistics $Z^TZ, Z^T X$.
        """
```

alphacsc/_base.py
```diff
@@ -0,0 +1,407 @@
...
+        self.max_iter = max_iter
+        self.momentum = momentum
+        self.rng = check_random_state(random_state)
+        self.verbose = verbose
+        self.debug = debug
+        self.D_init = D_init
+
+        self.init_strategy = get_init_strategy(
+            n_times_atom, self.get_D_shape(), self.rng, D_init
+        )
```
tomMoral, line 206: A classic convention is to use `check_init_strategy` instead of `get`.

alphacsc/convolutional_dictionary_learning.py
```diff
@@ -50,7 +50,9 @@
         Stopping criterion. If the cost descent after a uv and a z update is
         smaller than eps, return.
     reg : float
-        The regularization parameter.
+        The regularization parameter such that
+        :math:`\\lambda = \\text{{reg}} \\cdot \\sigma`,
+        with :math:`\\sigma` the standard deviation of the input signal X.
```
tomMoral, line 55: Is it really this? is it not `reg * lambda_max`?

alphacsc/convolutional_dictionary_learning.py
```diff
@@ -115,7 +118,7 @@
     .. math::
         \min_{D, Z} \sum_{n=1}^N
             \frac{1}{2} \|X^{(n)} - \sum_{k=1}^K D_k*Z^{(n)}_k\|_2^2
-            + \lambda\|Z^{(n)}\|_2
+            + \lambda\|Z^{(n)}\|_1
```
tomMoral, line 121: good catch!

alphacsc/learn_d_z_multi.py
```diff
@@ -239,6 +239,8 @@ def learn_d_z_multi(X, n_atoms, n_times_atom, n_iter=60, n_jobs=1,
         # Rescale the solution to match the given scale of the problem
         z_hat *= std_X
         reg = z_encoder.reg * std_X
+        if solver_z == 'no-overlap':
+            reg *= std_X
```
tomMoral, line 243: Why do you need to square `std_X` here?

alphacsc/convolutional_dictionary_learning.py
tomMoral: Oh actually this is confusing so good that you raised this up.
Formally, we normalize with `std` as this makes the algorithm much more stable.
But this means that we change the meaning of `reg`, which is surprising.
I think we should expand this comment (probably in a follow up PR).
Can you open an issue?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,750 @@
+import numpy as np
+import numba as nb
+import pyfftw
+
+from ._base import BaseZEncoder, BaseDSolver
+from .update_d_multi import prox_d
+
+# =================================
+# Helping functions for numba types
```
tomMoral, line 9:
```suggestion
# Helper functions for numba types
```

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,750 @@
...
+    nb.void(_float64_r(2), nb.int32, nb.float64,
+            _float64_w(), _int32_w(), _int32_w(), _float64_w()),
+    cache=True, nogil=True
+)
+def _dp_fft(proj, n_times_atom, reg,
+            dp, last, atom_index, atom_coeff):
+    """
+    Computes z for one signal X, using proj, an array such that
+    proj[i, t] is the dot product between D[i] and X[:, t-n_times_atom+1:t+1]
+
```
tomMoral, line 44:
```suggestion
    Computes z for one signal X, using dynamic programming.
    
    Parameters
    ----------------
    proj: ndarray, shape (n_atoms, n_times_valid)
        contains proj[i, t], the dot product between D[i] and X[:, t-n_times_atom+1:t+1].
    n_times_atom: int
        The length of an atom in the dictionary
    ....
```

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,750 @@
...
+    Notes
+    -----
+    Time Complexity: O(nnz * n_atoms * n_channels * n_times_atom),
+        where nnz is the number of non zero entries in z.
+    """
+    E0, Ereg = XtX, 0
+    N = X.shape[0]
+    p, C, L = D.shape
+    for trial in range(N):
```
tomMoral, line 157: Couldn't this be done in parallel?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,750 @@
...
+        self._compute_dense_z_hat()
+        return self.z_hat
+
+    def get_z_nnz(self):
+        self._compute_dense_z_hat()
+        return self.nnz_atom
+
+    def set_D(self, D):
+        self.D_hat = D
+        self._prox()
```
tomMoral, line 703: you need to reset the cost here no? otherwise `_update_z` won't return the right value?

tomMoral (commented):
Did a pass with some comments.
Could you also add an example on the gait data?

alphacsc/_base.py
Nanored4498: Done in 99e9f05

alphacsc/_base.py
Nanored4498: Yes, it is. I removed these parameters in the constructor.

alphacsc/convolutional_dictionary_learning.py
Nanored4498: As we introduce CSC-l0, we should maybe replace the l1 norm with a lp norm and explain that p depends on the ZEncoder.

tomMoral: Good point

alphacsc/convolutional_dictionary_learning.py
Nanored4498: I created the issue here #131

alphacsc/learn_d_z_multi.py
Nanored4498: It is because of the l0 norm.
When using l1 norm, multiplying $\| \sigma^{-1} X - D \ast z_\sigma \|_2^2 + \lambda \| z_\sigma \|_1$ by $\sigma^2$ gives $\| X - D \ast z \|_2^2 + \sigma \lambda \| z \|_1$, with $z = \sigma \cdot z_\sigma$.
While with l0 norm, multiplying $\| \sigma^{-1} X - D \ast z_\sigma \|_2^2 + \lambda \| z_\sigma \|_0$ by $\sigma^2$ gives $\| X - D \ast z \|_2^2 + \sigma^2 \lambda \| z \|_0$, as $\| z \|_0 = \| z_\sigma \|_0$

alphacsc/_no_overlap.py
Nanored4498: Yes, exact, self._prox() reset the cost ;)

alphacsc/learn_d_z_multi.py
tomMoral: Can you add a comment in the code?

Nanored4498: Done in 475320b

alphacsc/update_z_multi.py
```diff
@@ -198,10 +198,9 @@ def callback(xk):
             lbfgs_kwargs['factr'] = lbfgs_kwargs['tol'] / np.finfo(float).eps
         del lbfgs_kwargs['tol']
 
-        lbfgs_kwargs['disp'] = lbfgs_kwargs['verbose']
         del lbfgs_kwargs['verbose']
```
tomMoral: Does this means that `verbose` is deprecated?

doc/Makefile
```diff
@@ -51,6 +52,12 @@ html-noplot:
 	@echo
 	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
 
+.PHONY: html
+html-local:
+	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	@echo
+	@echo "Build finished. The HTML pages are in $(BUILDDIR)/html."
```
tomMoral, line 59: is it to avoid uing `mpiexec`? if yes I would do the reverse: rename `html` in `html-mpi` and have a local version by defautl.

examples/no-overlap/plot_no_overlap.py
```diff
@@ -0,0 +1,109 @@
...
+channel = 0
+
+fig_hat, ax_hat = plt.subplots()
+ax_hat.plot(X[trial, channel, 5000:5800],
+            label='right foot vertical acceleration (ORIGINAL)')
+ax_hat.plot(X_hat[trial, channel, 5000:5800],
+            label='right foot vertical acceleration (RECONSTRUCTED)')
+ax_hat.set_xlabel('time (x10ms)')
+ax_hat.set_ylabel('acceleration ($m.s^{-2}$)')
+ax_hat.legend()
```
tomMoral, line 119: Add `plt.show()` at the end, so when run as a python file, the result shows.

examples/no-overlap/plot_no_overlap.py
```diff
@@ -0,0 +1,109 @@
...
+
+X_hat = construct_X_multi(cdl._z_hat, cdl._D_hat)
+diff = X - X_hat
+E0 = .5 * diff.ravel() @ diff.ravel() / X.var()
+E = E0 + lmbd * nnz / X.var()
+print("l2-objective:", round(E0, 1))
+print("BatchCDL objective:", round(E, 1))
+
+###############################################################################
+# Plot a small part of the original and reconstructed signals
```
tomMoral, line 105: Can you show the learned dictionary?

Also can you show the difference with another algorithm?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+            _float64_w(), _int32_w(), _int32_w(), _float64_w()),
+    cache=True, nogil=True
+)
+def _dp_fft(proj, n_times_atom, reg,
+            dp, last, atom_index, atom_coeff):
+    """
+    Computes z for one signal X, using dynamic programming.
+    This function uses an array `proj` of dot products between D and X.
+    It then differs from `_dp_prod` which computes those dot products
+    on the fly.
```
tomMoral:
```suggestion
    Computes z for one signal X, using dynamic programming.
    
    This function uses an array `proj` of pre-computed dot products between D and X.
    It differs from `_dp_prod` which computes those dot products on the fly when the signals are small enough.
```

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+
+    Parameters
+    ----------------
```
tomMoral:
```suggestion
    Parameters
    ----------
```

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+
+@nb.njit(
+    nb.void(_float64_r(3), _float64_r(), _float64_r(2), nb.float64,
+            _float64_w(), _int32_w(), _int32_w(), _float64_w()),
+    cache=True, nogil=True
+)
+def _dp_prod(D, D_mul, X, reg,
+             dp, last, atom_index, atom_coeff):
+    """
+    Computes z for one signal X, using D, the dictionary.
```
tomMoral, line 106:
```suggestion
    Computes z for one signal X, using D, the dictionary.
    
    This function differs from _df_fft which used precomputed
    dot products using FFT.
```

Also, parameters are missing.

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+@nb.njit(
+    nb.int32(_float64_r(), _int32_r(), _int32_r(), _float64_r(), nb.int32,
+             _int32_w(2), _float64_w()),
+    cache=True, nogil=True
+)
+def _get_nz_values(D_mul, last, atom_index, atom_coeff, n_times_atom,
+                   nz_index, nz_coeff):
+    """
+    Retrieve z as a sparse representation from the array `last`,
+    obtained via dynamic programming (DP).
```
tomMoral, line 174: Precise what are the buffer and the one updated inplace.

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+               nb.float64, nb.float64,
+               _int32_w(3), _float64_w(2)),
+    cache=True, nogil=True
+)
+def _compute_z_from_T(D, D_mul, X, nnz,
+                      XtX, penalty,
+                      nz_index, nz_coeff):
+    """
+    Computes the best z value for a gien dictionary D and
+    a temporal support T stored in nz_index
```
tomMoral, line 262:
```suggestion
    a temporal support T stored in (nz_index, nz_coef)
```

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+    return .5 * E0 + penalty * Ereg
+
+
+@nb.njit(
+    nb.float64(_float64_r(3), _float64_r(3), _int32_r(), _int32_r(3),
+               nb.float64, nb.float64),
+    cache=True, nogil=True
+)
+def _compute_objective(D, X, nnz, nz_index,
+                       XtX, penalty):
```
tomMoral: Use the same argument order for all functions.
If possible, pass first array arguments, then parameters (reg/penalty) then buffers

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+    Find the patch with the maximum residual error.
+
+    Returns
+    -------
+    max_error_ind : int64
+        An integer such that:
+        * (max_error_ind >> 32) is the trial index in which the patch
+            with max error is found.
+        * (max_error_ind & 0xffffffff) is the start time of the patch
+            with max error.
```
tomMoral: why not return 2 ints?

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,753 @@
...
+        self._update_z()
+        return self.cost
+
+    def get_max_error_patch(self):
+        self._update_z()
+        ind = _find_max_error_patch(self.nnz, self.nz_index, self.nz_coeff,
+                                    self.D_hat, self.X)
+        atom_ind = ind >> 32
+        t = ind & ((1 << 32)-1)
+        return self.X[atom_ind, :, t:t+self.n_times_atom][None].copy()
```
tomMoral: This would feel more natural
```suggestion
        atom_ind, t = _find_max_error_patch(
            self.nnz, self.nz_index, self.nz_coeff, self.D_hat, self.X
        )
```

tomMoral (commented):
A few comment on the form but overall, we are close to done here :)

alphacsc/update_z_multi.py
Nanored4498: `disp` is deprecated and will be removed in scipy 1.18.

doc/Makefile
Nanored4498: Yes, it was to avoid using mpi. I renamed the rules/

alphacsc/_no_overlap.py
Nanored4498: I had not taken a look at numba tuples. Just replaced it with tuples ;)

alphacsc/_no_overlap.py
Nanored4498: It is only stored inside `nz_index` as `nz_coeff` stores the values of the non-zero entries of z. It is clarified in the new Parameters section.

alphacsc/convolutional_dictionary_learning.py
Nanored4498: Changed this point here :
https://github.com/alphacsc/alphacsc/pull/130/changes/1ab0e96647326164128521ead3cb7248ee7bffc6#diff-c2f514ef9971c1ad34c54ad15f52790b2e7c5c15313da81cefa469d3304ca8d6R121-R124

alphacsc/_no_overlap.py
Nanored4498: Yes, it is possible. Done using numba multi-threading in 1ab0e96.

examples/no-overlap/plot_no_overlap.py
Nanored4498: Some comparisons with other algorithms have been added here https://github.com/alphacsc/alphacsc/pull/130/changes#diff-62a47a741024871dae439122e021aabb5c1039fbfc8cc0baee5d536d422f595eR129-R168

alphacsc/learn_d_z_multi.py
```diff
@@ -164,10 +164,18 @@ def learn_d_z_multi(X, n_atoms, n_times_atom, n_iter=60, n_jobs=1,
     # initialization
     start = time.time()
 
+    # Use reg parameter for intialization with no-overlap
+    if (
+        isinstance(D_init, str) and D_init == "no-overlap"
+        and D_init == solver_d
+    ):
```
tomMoral, line 171: I would allow to use `no-overlap` anytime and remove the `D_init == solver_d` condition.

alphacsc/_no_overlap.py
```diff
@@ -751,3 +1076,44 @@ def update_D(self, z_encoder):
...
         return prox_d(D_hat)
+
+
+# ==============================================
+# Dictionary initializer for CSC without overlap
+# ==============================================
+
+class NoOverlapStrategy():
+    """A class that creates a dictionary for a specified shape with
+    a strategy thought for the no-overlap case.
```
tomMoral, line 1091: Add a comment to explain the algo.

alphacsc/init_dict.py
```diff
@@ -203,6 +211,11 @@ def init_dictionary(X, n_atoms, n_times_atom, uv_constraint='separate',
     elif D_init == 'greedy':
         raise NotImplementedError()
 
+    elif D_init == 'no-overlap':
+        assert not rank1
```
tomMoral: if rank1, do `D_hat = get_uv(D_hat)` after the init.

alphacsc/learn_d_z_multi.py
Nanored4498: This if statement is only for using the same `reg` parameter as the one used for the z-encoder when `reg` is not already defined in `init_kwargs`. As for other z-encoder `reg` has not the same meaning, I think it is better to raise an error stating that the `reg` parameter of the non-overlap initialization strategy is not defined in `init_kwargs`.

alphacsc/_no_overlap.py
Nanored4498: Done in 305d401

alphacsc/init_dict.py
Nanored4498: Done in 9add39e

alphacsc/_no_overlap.py
```diff
@@ -1085,6 +1089,9 @@ def prox(self, D_hat):
 class NoOverlapStrategy():
     """A class that creates a dictionary for a specified shape with
     a strategy thought for the no-overlap case.
+    This strategy first compute a temporal support minimizing the energy
+    E = - 0.5 * || X ||_2^2 + reg * || z ||_0
+    Then it applies the NoOverlapDSolver over the computed temporal support.
```
tomMoral: Simplify this comment

alphacsc/_no_overlap.py
```diff
@@ -0,0 +1,1126 @@
...
+
+    Parameters
+    ----------
+    shape: tuple
+        Expected shape of the dictionary. (n_atoms, n_channels, n_times_atom)
+    reg: float
+        Regularization parameter in the minimized energy
+        E = - 0.5 * || X ||_2^2 + reg * || z ||_0
```
tomMoral: ditto

alphacsc/_no_overlap.py
Nanored4498: Done in cbc6767

alphacsc/_no_overlap.py
Nanored4498: Done in cbc6767

tomMoral (approved):
LGTM! thx

PR #133, opened 2026-09-22
CI Change python version in CI

Nanored4498:
Update the versions of python used in the CI.


.github/workflows/unittests.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9, "3.10"]
+        python-version: [3.10, 3.12, 3.14]
```
tomMoral: Otherwise it revert to 3.1 🤣 
```suggestion
        python-version: ["3.10", 3.12, 3.14]
```

Nanored4498: oh yes thanks

